PR #8, opened 2015-02-18
Remove dependency on 'collection'

kaendfinger:
`args` was using `package:collection/wrappers.dart` for creating unmodifiable collection views, but this functionality is built directly into the SDK in `dart:collection`

This would be a breaking change, because of transitive dependencies.


nex3:
> This would be a breaking change, because of transitive dependencies.

We actually don't consider dependency changes to be breaking, since pub can figure out how to resolve them without user intervention.

PR #40, opened 2015-09-24
Make Command.argResults and Command.globalResults public to facilitate testing Command subclasses with mock ArgResults.

iansf:
This makes it possible to write Command tests that look like this:

```
import 'dart:async';

import 'package:args/args.dart';
import 'package:args/command_runner.dart';
import 'package:mockito/mockito.dart';
import 'package:test/test.dart';

class ExampleCommand extends Command {
  final name = 'install';
  final description = 'Amazing example!';
  ExampleCommand();

  @override
  Future<int> run() async {
    if (argResults['bad-option'] == true) {
      return 2;
    }

    if (argResults['carefully-checked-option'] == true) {
      return 0;
    }

    print('You need to specify an option to use $name!');
    return 1;
  }
}

class MockArgResults extends Mock implements ArgResults {}

main() => defineTests();

defineTests() {
  group('example', () {
    test('returns 0 when called with carefully-checked-option', () {
      ExampleCommand command = new ExampleCommand();

      MockArgResults results = new MockArgResults();
      when(results['carefully-checked-option']).thenReturn(true);
      when(results['bad-option']).thenReturn(false);
      command.argResults = results;
      command
          .run()
          .then((int code) => expect(code, equals(0)));
    });
  });
}
```

This seems like the natural way to test commands, rather than having to pass a List&lt;String&gt; to a CommandRunner.

What do you think?

### DIFF
--- a/lib/command_runner.dart
+++ b/lib/command_runner.dart
@@ -158,8 +158,8 @@ Run "$executableName help <command>" for more information about a command.''';
         // Step into the command.
         argResults = argResults.command;
         command = commands[argResults.name];
-        command._globalResults = topLevelResults;
-        command._argResults = argResults;
+        command.globalResults = topLevelResults;
+        command.argResults = argResults;
         commands = command._subcommands;
         commandString += " ${argResults.name}";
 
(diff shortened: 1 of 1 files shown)


nex3:
This is a lot of extra API surface—surface that adds extra mutability—for something that's already really easy to do just by passing strings. Mocking is a great fit for APIs that interact with the real world like networking and databases, but here you can already set the options the same way real-world code does just by passing in strings. Adding a mock just makes the test code more complex and less similar to reality.

iansf:
Thanks for the reply!  To make sure I understand your suggestion, is the following how you recommend testing Command subclasses?

```
defineTests() {
  group('example', () {
    test('returns 0 when called with carefully-checked-option', () {
      ExampleCommand command = new ExampleCommand();

      new CommandRunner('example_app', '')
          ..addCommand(command)
          ..run(['example', '--carefully-checked-option'])
          .then((int code) => expect(code, equals(0)));
    });
  });
}
```

nex3:
I personally usually write more integration-style tests for command-line apps; I want to make sure they work end-to-end, so I start them as processes. But that involves a lot of overhead, and if you didn't want to incur it then I would recommend something like that. I'd probably wrap it up in a helper method, though.

iansf:
Ok -- thanks for your perspective!  Doing it this way, or as a full integration test, breaks with how I would prefer to write my unit tests, but I agree that this approach is not exercising all that much more code than my preferred approach, so it is not unreasonable.  Thanks again for your help!

PR #44, opened 2016-02-02
Fix #39 with docs change

### LINKED ISSUE #39: README: options instead of argResults
In the README file, in the CommandRunner example, it reference the (old?) non-existing 'options' member, instead of the correct 'argResults' member. The example need to get updated.

srawlins:


nex3:
Thanks for the pathces!

PR #45, opened 2016-02-06
Document allowTrailingOptions in code

kasperpeulen:


sethladd:
cool, thanks!

cc @munificent

zoechi:
This isn't an improvement IMHO. Now when `null` is passed this breaks.

kasperpeulen:
@zoechi Why would you pass `null`? I just like to see the default value here:

image

zoechi:
I didn't say I want to pass `null`. What I say is that when passing `null`, the previous behavior was like `false` was passed but now it will probably throw eventually. That's a breaking change for no reason. You can still add a default value but the lines you removed below break the function.

kasperpeulen:
@zoechi I agree, I changed those lines back.

nex3:
Thanks for the patch!

PR #55, opened 2016-10-24
Print command help for "bin --help command"

### LINKED ISSUE #54: `bin.dart --help foo` – should print help for foo, not run foo
Just ran into this

nex3:
Closes #54

### DIFF
--- a/lib/command_runner.dart
+++ b/lib/command_runner.dart
@@ -220,8 +220,8 @@ abstract class Command {
 
     var invocation = parents.reversed.join(" ");
     return _subcommands.isNotEmpty
-    ? "$invocation <subcommand> [arguments]"
-    : "$invocation [arguments]";
+        ? "$invocation <subcommand> [arguments]"
+        : "$invocation [arguments]";
   }
 
   /// The command's parent command, if this is a subcommand.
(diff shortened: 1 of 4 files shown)


nex3:
cc @kevmoo

lib/command_runner.dart
```diff
@@ -217,8 +220,8 @@ abstract class Command {
 
     var invocation = parents.reversed.join(" ");
     return _subcommands.isNotEmpty
-        ? "$invocation <subcommand> [arguments]"
-        : "$invocation [arguments]";
+    ? "$invocation <subcommand> [arguments]"
+    : "$invocation [arguments]";
```
munificent: What happened to the indentation here and below?

lib/command_runner.dart
```diff
@@ -132,56 +132,59 @@ Run "$executableName help <command>" for more information about a command.''';
...
+      return;
+    }
+
+    // Make sure there aren't unexpected arguments.
+    if (!command.takesArguments && argResults.rest.isNotEmpty) {
+      command.usageException(
+          'Command "${argResults.name}" does not take any arguments.');
+    }
+
+    await command.run();
```
Hixie: this is a breaking change -- it means the command's exit code is no longer propagated.
See https://github.com/dart-lang/args/issues/58

PR #57, opened 2016-11-08
Added missing return

algobardo:


kevmoo:
@algobardo Are you using the value returned from the command?

@nex3 Reasonable?

algobardo:
I'm using the returned value.
Why do you need to await for the command to run.
Isn't 
```
return command.run();
```
an even better solution ?

PR #60, opened 2016-12-06
Use arrow functions

mohamedhayibor:
Wasn't so sure whether to indent 4 or 2 spaces after the arrow. @nex3 

Also, let me know if a `table of contents` or `contents` in short, would be useful in the Readme.


kevmoo:
I'll let @nex3 comment – but my feeling is this is going a bit too far. I'm a fan of arrows for functions – as long as they are pretty trivial.

mohamedhayibor:
Thanks for the feedback @kevmoo. You can close the pr. I will try helping out in other issues.

PR #64, opened 2017-02-15
Add support for testing in the VM on travis.

eseidelGoogle:
Unclear if this adds value. My thinking is trying to make this
(and eventually other) dart packages feel more appealing to
the casual non-google contributor in part by helping to know
if the package is working for non-google configs.

I've run into the trouble with public Github Dart packages not
working for non-google configs before, and my thought is
Travis might be one way to help with that.

Thoughts welcome!  Happy to discard this if not helpful.

Someone with admin powers for dart-lang (like @kevmoo)
would have to flip the bit at:
https://travis-ci.org/dart-lang/args
to actually turn on testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: dart
+dart:
+  - dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
 language: dart
 dart:
   - dev
+  - stable
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![Build Status](https://travis-ci.org/dart-lang/args.svg?branch=master)](https://travis-ci.org/dart-lang/args)
+
 Parses raw command-line arguments into a set of options and values.
 
 This library supports [GNU][] and [POSIX][] style options, and it works


kevmoo:
@eseidelGoogle wanna do a force-push to your branch? Just to kick travis?

.travis.yml
```diff
@@ -0,0 +1,3 @@
+language: dart
+dart:
+  - dev
```
kevmoo: testing on stable is good, too.

Also: I encourage folks to include at least the current stable branch explicitly here.

...keeps us honest about actually supporting previous SDKS https://github.com/dart-lang/shelf/blob/master/.travis.yml

eseidelGoogle:
Fixed.  Looks like it's kicking travis now.  Please feel free to squash-and-merge if this meets expectations :D

.travis.yml
kevmoo: Would you at least add `1.22.0` explicitly as well?

nex3:
I'd like to hold off on Travis-ifying our packages until I have a chance to improve Travis's built-in  support for Dart and for the test package.

kevmoo:
I'm assuming any work to redo travis won't break existing users.

Given that this config is pretty standard and that it improves the experience for folks sending PRs, I see zero reason *not* to take this now.

nex3:
Ideally it won't break existing users too badly, but I'm not sure how much we can guarantee. I also don't want to have to worry about how to fix packages if they do run into infrastructure issues of any kind in the future, without a sound basis on which to do so.

matanlurey:
@nex3 Do we have a tracking bug for fixing Travis?

Some contributors have mentioned they don't feel comfortable contributing to packages that don't have Travis enabled, as they aren't able to ensure they didn't break anything. In packages where we don't have Travis (like HTTP - see https://github.com/dart-lang/http/pull/58) we frequently break things for users.

I use Travis in about a dozen packages successfully, even if support is non-perfect. Can we enable this and just make it better instead of blocking incremental improvements to code safety?

nex3:
> Do we have a tracking bug for fixing Travis?

We don't really have a repo that would make sense for a tracking issue.

I agree that getting Travis up and running is important—I feel the pain of not having CI for packages more than just about anyone. But I expect many packages to require substantially more or different configuration than the default and I don't want to have to invest that effort for each package and then have to change all their configurations again when the infrastructure gets up to snuff.

matanlurey:
Sorry, I'm not convinced. Specifically, `args` is a super simple package. Adding travis support to help external contributors doesn't seem risky in anyway, and it's already done if we merge this PR. Can we make the travis support better? Sure.

eseidelGoogle:
So I definitely didn't mean to start any trouble. :/  Well no more than normal at least.

If I wrote a different PR to use the (older) dart shell-script style like:
https://github.com/google/vector_math.dart/blob/master/tool/travis.sh
would that be accepted?

I'm really happy to do whatever.  In part I was looking for a google dart repository to improve as well as test out our best-practices travis and dart-coveralls support (I'm separately working with the new maintainer of dart-coveralls to fix it and release a new version compatible with the latest sdks -- I obviously can also use my own personal repos for any ad-hoc testing and will as necessary.)

.travis.yml
eseidelGoogle: I'm happy to add anything you like, I used the named versions to avoid things going stale. :/  Doesn't "dev" cover 1.22?

eseidelGoogle:
I'm also happy to wait, btw.  Hacking on dart stuff is just fun free-time coding for me. :)

eseidelGoogle:
Upon further reflection, I really am just causing noise here.  We can have a grand travisification at a future time instead of worrying about a one-off here. Everyone involved in this thread is working on much more important things for our work priorities. :)  Closing.

PR #67, opened 2017-04-07
Enable Travis-CI

kevmoo:
CC @eseidelGoogle

### DIFF
--- a/example/test_runner.dart
+++ b/example/test_runner.dart
@@ -20,49 +20,53 @@ main() {
       abbr: 'c',
       defaultsTo: 'none',
       help: 'Specify any compilation step (if needed).',
-      allowed: ['none', 'dart2js', 'dartc'],
+      allowed: [
+        'none',
+        'dart2js',
+        'dartc'
+      ],
       allowedHelp: {
-    'none': 'Do not compile the Dart code (run native Dart code on the'
-        ' VM).\n(only valid with the following runtimes: vm, drt)',
-    'dart2js': 'Compile dart code to JavaScript by running dart2js.\n'
-        '(only valid with the following runtimes: d8, drt, chrome\n'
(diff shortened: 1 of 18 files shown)


example/test_runner.dart
```diff
@@ -20,49 +20,53 @@ main() {
       abbr: 'c',
       defaultsTo: 'none',
       help: 'Specify any compilation step (if needed).',
-      allowed: ['none', 'dart2js', 'dartc'],
+      allowed: [
+        'none',
+        'dart2js',
+        'dartc'
```
eseidelGoogle: Is this the new dartfmt behavior?  I'm surprised it doesn't require a trailing , to explode this 3-item list like this.

eseidelGoogle:
If I were writing this change I would have split out the dartfmt application from the travisification. :/

But lgtm.  Honestly you should get someone who actually codes-fulltime to review this though.  e.g. @munificent or @lrhn

kevmoo:
> If I were writing this change I would have split out the dartfmt application from the travisification

@eseidelGoogle It's two commits. I'll land it as two separate commits. 😄

eseidelGoogle:
I guess my Github-fu is too weak to view it as such. :D 

Regardless, it looks fine.  Bob may have more useful commentary for you. :p

example/test_runner.dart
munificent: dartfmt's behavior can be subtle in some cases. If I recall, there is a rule that says that if all of the trailing arguments of an argument lists are collection literals, then if any of them split (like the allowedHelp map below), then they all should. It avoids weird cases like:

```dart
function(first: [
  "stuff",
  "more",
  "stuff",
  "more",
  "stuff",
  "more",
  ], second: ["short"], third: [
  "stuff",
  "more",
  "stuff",
  "more",
  "stuff",
  "more",
])
```

PR #71, opened 2017-05-16
Add ArgParser.allowAnything().

### LINKED ISSUE #70: No way for an ArgParser to accept any arguments
I want to write a subcommand that takes all its arguments and passes them unmodified through to another process. There doesn't seem to be a way to turn off ArgParser's error handling.

nex3:
Closes #70

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 0.13.8
 
-* Add `new ArgParser.allowAnything()`. This allow any input, without parsing any
-  options.
+* Add `new ArgParser.allowAnything()`. This allows any input, without parsing
+  any options.
 
 ## 0.13.7
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.13.8
+
+* Add `new ArgParser.allowAnything()`. This allow any input, without parsing any
+  options.
+
 ## 0.13.7
 
 * Add explicit support for forwarding the value returned by `Command.run()` to
--- a/lib/src/allow_anything_parser.dart
+++ b/lib/src/allow_anything_parser.dart
@@ -1,0 +1,65 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'arg_parser.dart';
+import 'arg_results.dart';
+import 'option.dart';
+import 'parser.dart';
+
+/// An ArgParser that treats *all input* as non-option arguments.
+class AllowAnythingParser implements ArgParser {
+  Map<String, Option> get options => const {};
+  Map<String, ArgParser> get commands => const {};
+  bool get allowTrailingOptions => false;
+  bool get allowsAnything => true;
+
+  ArgParser addCommand(String name, [ArgParser parser]) {
+    throw new UnsupportedError(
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.13.8
+
+* Add `new ArgParser.allowAnything()`. This allow any input, without parsing any
```
munificent: "allow" -> "allows"

lib/src/allow_anything_parser.dart
```diff
@@ -0,0 +1,65 @@
...
+        "ArgParser.allowAnything().addOption() isn't supported.");
+  }
+
+  void addSeparator(String text) {
+    throw new UnsupportedError(
+        "ArgParser.allowAnything().addSeparator() isn't supported.");
+  }
+
+  ArgResults parse(List<String> args) =>
+      new Parser(null, this, args.toList()).parse();
```
munificent: Going through `Parser.parse()` doesn't seem super useful.

How about just returning the results directly here?

lib/src/arg_parser.dart
```diff
@@ -29,15 +30,26 @@ class ArgParser {
...
-          allowTrailingOptions: allowTrailingOptions);
+      new ArgParser._({}, {}, allowTrailingOptions: allowTrailingOptions);
+
+  /// Creates a new ArgParser that treats *all input* as non-option arguments.
+  ///
+  /// This is intended to allow arguments to be passed through to child
+  /// processes without needing to be redefined in the parent.
+  ///
+  /// Options may not be defined for this parser.
+  factory ArgParser.allowAnything() = AllowAnythingParser;
```
munificent: Instead of exposing a constructor for a subclass that basically says all of its methods can't be used, how about adding `addUnparsedCommand(String name)` to ArgParser. That would implicitly create an AllowAnythingParser under the hood but not give access to it.

CHANGELOG.md
nex3: Done.

lib/src/allow_anything_parser.dart
nex3: We need to have the logic in `Parser` anyway to support sub-commands that allow anything, and because the `newArgResults()` call for `allowAnything` is kinda complex, I thought it would be better to have a single canonical place for it.

lib/src/arg_parser.dart
nex3: I'm not sure how that would work with the command runner—the user is likely to still want a custom `Command` with a custom `run()` method, which means they need *some* implementation for `Command.argParser`.

More broadly, I think it's useful to give users the ability to slot in a parser that allows anything to any API that expects an `ArgParser`. `CommandRunner` is the most obvious example of such an API right now, but we should design for the possibility that more appear later.

munificent:
This still needs a test, then 👍

test/allow_anything_test.dart
```diff
@@ -0,0 +1,61 @@
...
+import 'package:args/args.dart';
+
+import 'utils.dart';
+
+void main() {
+  group('new ArgParser.allowAnything()', () {
+  ArgParser parser;
+  setUp(() {
+    parser = new ArgParser.allowAnything();
+  });
```
munificent: Nit: The indentation is wrong here.

nex3: I fixed that in 2d068df5280d919b0dfb37896b9ba4e0496d5f3b.

Issue #68, opened 2017-05-11
addSubCommand should have a "hide" option


Hixie:
I wish you could have hidden subcommands that didn't appear in the help, the same way you can have hidden arguments and options.

nex3:
You can do this today, actually. `Command` has a `hidden` getter that you can override in your custom commands.

PR #72, opened 2017-05-17
Document that Command.argParser can be overridden.

nex3:
See #68


nex3:
cc @Hixie

Hixie:
thanks!

presumably it's not ok to return multiple instances over time from this getter?

nex3:
That's right.

PR #75, opened 2017-07-14
Update .travis.yml: allows failures in dartfmt dev

kevmoo:


nex3:
Why are you using `allow_failures` here instead of just fixing the formatting failures?

kevmoo:
Because stable and dev disagree. Pick the release that will have less churn – for now.

nex3:
Running a test with `allow_failures` just consumes bot resources without actually providing any value. We should only be testing against one formatter version anyway—probably `dev` so we pick up the latest and greatest formatting.

PR #87, opened 2018-01-26
Type Command.run()'s return value as FutureOr<T>

### LINKED ISSUE #85: Command.run should return a FutureOr<T>
Related to https://github.com/dart-lang/args/issues/76, but I think this actually does make sense since the return type is actually a FutureOr already, its just not expressed as such.

nex3:
Closes #85


nex3:
In the package world, documented contracts are considered binding as far as versioning goes, even if they aren't enforced. That's the only way to keep versioning sane in cases where the language isn't strict enough to enforce constraints.

PR #93, opened 2018-02-21
Add a separate addMultiOption() method

### LINKED ISSUE #90: Impossible to set default value for option with splitComma
An option with allowMultiple and splitComma set allows a comma-separated list to be parsed into a list of multiple values. For example:

```dart
final results = new ArgParser().addOption(
    'foo',
    allowMultiple: true,
    splitCommas: true,
).parse('a,b,c');
print(results['foo']);
// [ 'a' , 'b' , 'c' ]
```

However there's no possible way to set the default value. `addOption` accepts a `defaultsTo` parameter, however this is declared to be a String so passing a list fails at compile time:

```dart
// ERROR: The argument type 'List<String>' can't be assigned to the parameter type 'String'. #argument_type_not_assignable
```

But a string argument won't be split, even if it's comma-separated, and will just return the entire string in a single-element list:

```dart
final results = new ArgParser().addOption(
    'foo',
    defaultsTo: const ['a', 'b'].join(',').toList(),
    allowMultiple: true,
    splitCommas: true,
).parse('');
print(results['foo']);
// [ 'a,b' ]
```

nex3:
This allows us to be more specific in our types and to keep flags that are
specific to multi-options in the place where they make sense.

Closes #90

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ dart_task:
 
 matrix:
   include:
-  - dart: stable
+  - dart: dev
     dart_task: dartfmt
 
 # Only building master means that we don't run two builds for each pull request.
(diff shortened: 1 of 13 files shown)


nex3:
cc @Hixie who uses args and has strong opinions about things

PR #97, opened 2018-04-09
Added optional `allowTrailingOptions` to `Command` constructor

kevmoo:


nex3:
The `Command` class is meant to be customized via inheritance. The recommended way of customizing `allowTrailingOptions` is to override the `argParser` getter.

PR #98, opened 2018-04-09
Remove lints

nex3:
Many of these aren't passing, and I don't believe in supporting lints that are generally unmaintained and non-canonical.


matanlurey:
Looks like @kevmoo is a better reviewer here:
https://github.com/dart-lang/args/commit/a27101e45aaacb28419e48cd82e27aaae578c532#diff-d62227308ce5fdb544ee6b0a8b157a69

PR #104, opened 2018-06-10
Update README to use addMultiOption

srawlins:


srawlins:
Done. I don't think its worth a changelog entry.

PR #105, opened 2018-07-11
Add optional wrapping of text for usage messages.

### LINKED ISSUE #41: Help output should be line-wrapped to screen width

gspencergoog:
This adds an _optional_ `maxLineLength` argument to the ArgParser that will automatically wrap help text at the given column on the terminal.

Fixes #41

### DIFF
--- a/lib/src/usage.dart
+++ b/lib/src/usage.dart
@@ -51,7 +51,12 @@ class Usage {
   /// content.
   int newlinesNeeded = 0;
 
-  Usage(this.optionsAndSeparators);
+  /// The horizontal character position at which help text is wrapped. Help that
+  /// extends past this column will be wrapped at the nearest whitespace (or
+  /// truncated if there is no available whitespace).
+  final int maxLineLength;
+
+  Usage(this.optionsAndSeparators, {this.maxLineLength});
 
   /// Generates a string displaying usage information for the defined options.
   /// This is basically the help text shown on the command line.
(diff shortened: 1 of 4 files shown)


gspencergoog:
cc: @nex3 @munificent

lib/src/usage.dart
```diff
@@ -171,8 +176,52 @@ class Usage {
     numHelpLines = 0;
   }
 
+  /// Wrap a single line of text into lines no longer than length.
+  /// Try to split at whitespace, but if that's not good enough to keep it
+  /// under the length, then split in the middle of a word.
+  List<String> wrap(String text, int length) {
```
munificent: This is only called by `write()`, right?

How about:

* Make it private.
* Move `const int minColumnWidth = 10;` and the width calculation into here instead of passing it in.

lib/src/usage.dart
```diff
@@ -171,8 +176,52 @@ class Usage {
...
+      }
+    }
+    result.add(text.substring(currentLineStart));
+    return result;
+  }
+
   void write(int column, String text) {
+    const int minColumnWidth = 10;
     var lines = text.split('\n');
+    if (column == columnWidths.length && maxLineLength != null) {
```
munificent: The `column == columnWidths.length` part isn't super obvious. How about adding:

```
// If we are writing the last column, word wrap it to fit.
```

lib/src/usage.dart
```diff
@@ -261,3 +310,20 @@ String padRight(String source, int length) {
 
   return result.toString();
 }
+
+bool isWhitespace(String text, int index) {
+  final int rune = text.codeUnitAt(index);
+  return ((rune >= 0x0009 && rune <= 0x000D) ||
```
munificent: Nit: None of these parens are needed. If you prefer the ones around the `&&`, that's OK. But the outermost ones are definitely unnecessary.

lib/src/usage.dart
```diff
@@ -171,8 +176,52 @@ class Usage {
...
+    return result;
+  }
+
   void write(int column, String text) {
+    const int minColumnWidth = 10;
     var lines = text.split('\n');
+    if (column == columnWidths.length && maxLineLength != null) {
+      var wrappedLines = <String>[];
+      var start = 0;
+      columnWidths.sublist(0, column).forEach((int width) => start += width);
```
munificent: I think this is a more idiomatic using `reduce()`. Or just a plain for-in loop.

test/usage_test.dart
```diff
@@ -314,6 +314,102 @@ void main() {
...
+      parser.addFlag('longNewline',
+          help:
+              'The flag with a really long help text and newlines\n\nthat will '
+              'still be wrapped because it is really long.');
+      parser.addFlag('solid',
+          help:
+              'The-flag-with-no-whitespace-that-will-be-wrapped-by-splitting-a-word.');
+      parser.addFlag('small1', help: ' a ');
+      parser.addFlag('small2', help: ' a');
+      parser.addFlag('small3', help: 'a ');
```
munificent: Nice tests!

Is it worth adding a test for how a line containing leading whitespace *and* word-wrapping is handled?

lib/src/usage.dart
gspencergoog: Good suggestions. Done!

lib/src/usage.dart
gspencergoog: Done.

lib/src/usage.dart
gspencergoog: OK.  Switched to reduce().

lib/src/usage.dart
gspencergoog: Removed the parens.

test/usage_test.dart
gspencergoog: Sure.  Added, along with one that has trailing whitespace.

gspencergoog:
Superseded by https://github.com/dart-lang/args/pull/107

PR #108, opened 2018-08-06
Add support for wrapping command descriptions and other help text.

gspencergoog:
When I started to use the wrapping feature, I realized that it only is wrapping the arguments, and really, it needs to wrap everything, especially the command descriptions in the command list.

This fixes things to make it so when the `usageLineLength` is set, then the output is all wrapped at that column.  The wrapping preserves indents properly, and can even do a hanging indent for the case where it prints the usage.

Output is unchanged when `usageLineLength` is not set.

Because it's no longer limited to the `Usage` class, I moved the wrapping logic to the utils file, and made it public within the package (but not exported). I added a separate set of tests for the utils file.


gspencergoog:
cc @munificent

lib/command_runner.dart
```diff
@@ -49,17 +49,20 @@ class CommandRunner<T> {
...
 
-Global options:
+${_wrap('Global options:')}
 ${argParser.usage}
 
-${_getCommandUsage(_commands)}
+${_getCommandUsage(_commands, lineLength: argParser.usageLineLength)}
 
-Run "$executableName help <command>" for more information about a command.''';
+${_wrap('Run "$executableName help <command>" for more information about a command.')}''';
```
munificent: There's enough interpolation and stuff going on in this literal that it's probably worth switching to a StringBuffer and building it up imperatively.

lib/command_runner.dart
```diff
@@ -49,17 +49,20 @@ class CommandRunner<T> {
 
   /// Returns [usage] with [description] removed from the beginning.
   String get _usageWithoutDescription {
+    const usagePrefix = "Usage:";
```
munificent: Nit: In the args package, we don't bother with `const` or `final` for locals. Functions are small enough that it's not worth the effort.

lib/command_runner.dart
```diff
@@ -376,15 +396,15 @@ abstract class Command<T> {
 
   /// Throws a [UsageException] with [message].
   void usageException(String message) =>
-      throw new UsageException(message, _usageWithoutDescription);
+      throw new UsageException(_wrap(message), _usageWithoutDescription);
```
munificent: I'm iffy about wrapping the message in the exception text. Exceptions are often caught programmatically. If you were to, say, log the message text to a file, you probably don't want it arbitrarily line-wrapped.

Is it possible to defer the wrapping until the exception is displayed instead of doing it eagerly here?

lib/command_runner.dart
```diff
@@ -400,13 +420,15 @@ String _getCommandUsage(Map<String, Command> commands,
   var buffer =
       new StringBuffer('Available ${isSubcommand ? "sub" : ""}commands:');
   for (var name in names) {
-    var lines = commands[name].summary.split("\n");
+    var columnStart = length + 5;
```
munificent: Nit: This can be hoisted out of the loop.

I would tend to do this not because it helps performance (it probably doesn't) but to make it more visually clear that the value of this variable doesn't depend on `name`.

lib/src/utils.dart
```diff
@@ -1,7 +1,147 @@
...
 String padRight(String source, int length) =>
     source + ' ' * (length - source.length);
+
+/// Wraps a block of text into lines no longer than [length].
+///
+/// Tries to split at whitespace, but if that's not good enough to keep it
+/// under the limit, then it splits in the middle of a word.
+///
+/// Preserves indentation (leading whitespace) for each line (delimited by '\n')
+/// in the input, and will indent wrapped lines the same amount.
```
munificent: "will indent" -> "indents".

lib/src/utils.dart
```diff
@@ -1,7 +1,147 @@
...
+///
+/// If [hangingIndent] is supplied, then that many spaces will be added to each
```
munificent: "will be" -> "are".

lib/src/utils.dart
```diff
@@ -1,7 +1,147 @@
...
+        ));
+      }
+    } else {
+      notIndented = wrapTextAsLines(
+        trimmedText,
+        length: length - leadingWhitespace.length,
+      );
+    }
+    String hangingIndentString;
+    result.addAll(notIndented.map<String>(
```
munificent: "<String>" isn't needed.

https://www.dartlang.org/guides/language/effective-dart/design#avoid-redundant-type-arguments-on-generic-invocations

lib/src/utils.dart
```diff
@@ -1,7 +1,147 @@
...
+      (String line) {
```
munificent: The type annotation isn't needed.

https://www.dartlang.org/guides/language/effective-dart/design#avoid-annotating-inferred-parameter-types-on-function-expressions

Also, can you dartfmt this?

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: args
-version: 1.5.0
+version: 1.5.1
```
munificent: "1.5.1-dev"

lib/command_runner.dart
munificent: OK, I'm convinced. Let's do it the way you suggest.

lib/src/utils.dart
```diff
@@ -39,18 +39,17 @@ String wrapText(String text, {int length, int hangingIndent}) {
...
+    var leadingWhitespace = line.substring(0, line.length - trimmedText.length);
+    List<String> notIndented;
     if (hangingIndent != 0) {
       // When we have a hanging indent, we want to wrap the first line at one
       // width, and the rest at another (offset by hangingIndent), so we wrap
       // them twice and recombine.
-      final firstLineWrap = wrapTextAsLines(
+      var firstLineWrap = wrapTextAsLines(
         trimmedText,
         length: length - leadingWhitespace.length,
```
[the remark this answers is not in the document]
gspencergoog: Done.

gspencergoog:
Done.  No, I don't have permission to merge.

kevmoo:
@munificent – wanna squash/merge?

gspencergoog:
Thanks for committing this!  Can you give me an idea of when it will be published, so Flutter can use it?

PR #111, opened 2018-12-17
Enable and fix a number of standard lints and deprecated member use

kevmoo:
Improves the package health when analyzed on the pub site

### DIFF
--- a/lib/src/arg_parser.dart
+++ b/lib/src/arg_parser.dart
@@ -2,6 +2,10 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+// TODO(nweiz): Remove this ignore when sdk#30084 is fixed or when args no
+// longer refers to its own deprecated members.
+// ignore_for_file: deprecated_member_use
+
 import 'dart:collection';
 
 import 'allow_anything_parser.dart';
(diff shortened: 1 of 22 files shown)


kevmoo:
@munificent trying to cleanup https://pub.dartlang.org/api/packages/args/metrics?reports

lib/src/arg_parser.dart
```diff
@@ -2,6 +2,10 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+// TODO(nweiz): Remove this ignore when sdk#30084 is fixed or when args no
```
munificent: "kevmoo", unless you're unilaterally signing up Natalie for work. :)

test/args_test.dart
```diff
@@ -112,15 +113,20 @@ void main() {
     test(
         'throws ArgumentError if splitCommas is passed with allowMultiple: '
         'false', () {
-      var parser = new ArgParser();
-      throwsIllegalArg(() => parser.addOption('flummox', splitCommas: true));
-      throwsIllegalArg(() => parser.addOption('flummox', splitCommas: false));
+      var parser = ArgParser();
+      throwsIllegalArg(() => parser.addOption(
```
munificent: Now that there's a line comment in there, I think this is clearer as a `{ ... }` lambda. Below too.

lib/src/arg_parser.dart
kevmoo: This was a copy-paste from analysis_options, so I left the owner as-is.

test/args_test.dart
kevmoo: ack

PR #117, opened 2019-05-31
Fix deprecated URLs

kevmoo:
Remove superfluous codereview.settings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -215,7 +215,7 @@
 ## 0.12.0+1
 
 * Remove the documentation link from the pubspec so this is linked to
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -215,7 +215,7 @@
 ## 0.12.0+1
 
 * Remove the documentation link from the pubspec so this is linked to
-  pub.dartlang.org by default.
+  pub.dev by default.
```
natebosch: Why change this retroactively? At the time of the change this was true

kevmoo: repo_manager does a search for outdated strings. This keeps it out of my filter.

PR #119, opened 2019-07-09
Adds --all to the help command

zanderso:
This PR adds flags to the omnipresent 'help' command to control output of all help text. It does this by adding an `--all` flag to the `help` command, and flags `--split` and `--output` that control where the output goes. The intention is that command line tools can e.g. auto-generate their documentation.

### DIFF
--- a/lib/src/help_command.dart
+++ b/lib/src/help_command.dart
@@ -2,12 +2,35 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
+import 'dart:io';
+
 import '../command_runner.dart';
+import 'utils.dart';
 
 /// The built-in help command that's added to every [CommandRunner].
 ///
 /// This command displays help information for the various subcommands.
 class HelpCommand<T> extends Command<T> {
+  HelpCommand() {
+    argParser.addFlag('all',
+        abbr: 'a',
+        defaultsTo: false,
+        negatable: false,
+        help: 'Output help for every command and subcommand.');
+    argParser.addOption('output',
+        abbr: 'o',
+        help: 'When --split is given, the output directory. '
+            'When --split is not given, the output file.',
+        valueHelp: 'OUTPUT');
+    argParser.addFlag('split',
+        abbr: 's',
+        defaultsTo: false,
+        negatable: false,
+        help:
+            'Split help output by subcommand into files written to --output.');
+  }
+
   @override
   final name = "help";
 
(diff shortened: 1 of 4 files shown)


kevmoo:
Asking @natebosch to look at this.

`--all` I get. Auto-adding a flag to every command runner that writes to disk seems a stretch.

natebosch:
Definitely a strong -1 to adding an import to `dart:io`. I'd rather see the split and output stuff handled externally

lib/src/help_command.dart
```diff
@@ -55,4 +84,67 @@ class HelpCommand<T> extends Command<T> {
...
+    } finally {
+      if (outSink != stdout) {
+        await outSink?.flush();
+        await outSink?.close();
+      }
+    }
+  }
+
+  String _subcommandChain(Command command) {
+    List<String> parents = <String>[command.name];
```
natebosch: `var` or `final` here instead of `List<String>`

lib/src/help_command.dart
```diff
@@ -55,4 +84,67 @@ class HelpCommand<T> extends Command<T> {
...
+    for (Command c = command.parent; c != null; c = c.parent) {
+      parents.add(c.name);
+    }
+    return parents.reversed.join("_");
```
natebosch: [nit] rest of the file is using single quotes - can we be consistent?

zanderso:
Can you suggest how it could be implemented externally with the same UI, that is passing flags to the built-in help command. My hope was that users of this wouldn't have to essentially re-implement a lot of the package to do something like this.

lib/src/help_command.dart
```diff
@@ -55,4 +84,67 @@ class HelpCommand<T> extends Command<T> {
...
+
+  // Walks the subcommand tree, yielding all help text.
+  Iterable<List<String>> get _helpMessages sync* {
+    yield <String>[runner.executableName, '${runner.usage}\n'];
+    final Iterable<MapEntry<String, Command>> topLevelCommands =
+        runner.commands?.entries;
+    if (topLevelCommands == null) {
+      return;
+    }
+    final List<MapEntry<String, Command>> commandStack =
```
natebosch: omit types on the left throughout.

zanderso:
I would also like to understand the downsides of importing `dart:io` into a package intended for use in command-line programs.

lib/src/help_command.dart
```diff
@@ -55,4 +84,67 @@ class HelpCommand<T> extends Command<T> {
...
+        List<MapEntry<String, Command>>.from(topLevelCommands);
```
natebosch: How about a `Queue` here? Would probably end up with then `command.value.subcommands.entries.forEach(commandStack.addFirst)`

zanderso:
`--all` by itself without the ability to write to files isn't going to get me where I'm going, so I'm going to close this and explore other approaches.

natebosch:
> I would also like to understand the downsides of importing `dart:io` into a package intended for use in command-line programs.

Nothing is stopping someone from using it on the web today and this would be breaking if anyone was doing it. That said, I can't find any examples so it could be that no one would be impacted.

In any case - it isn't an API that makes sense to me. It's also not something that we'd expect to be useful to end users of the binary - mostly the the author.

`--all` along with something like `csplit` with a pattern of `Command:` should get you pretty close I'd think...

PR #127, opened 2019-12-17
Updated README.md

kiesman99:
The usage of the CommandRunner was a bit unclear, so I made a suggestion for a more clarified implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,10 +273,23 @@ support for dispatching to [Command][]s based on command-line arguments, as well
 as handling `--help` flags and invalid arguments. For example:
 
 ```dart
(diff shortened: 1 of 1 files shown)


bsutton:
I just came here to raise an issue about the poor doco around CommandRunner, so it was a nice surprise to see that someone else had beaten me to the punch.

However I'm still unclear on how to add multiple top level commands.

e.g.
I have three commands:
start
renew
help

The CommandRunner appears to require a 'top' level command e.g.

git start
git renew
git help

So if there is a way to do this with the CommandRunning then can we have the doco changed to reflect this as well.

PR #128, opened 2020-01-31
Fix performance issues due to list.removeAt(0), which makes parsing O(n^2).

davidmorgan:
Add a test that checks that parse time does not increase more than 30x when input size grows from 10k to 100k. The current increase is about x2, so this is a very loose check that should be reliably green. It will only trigger if we hit O(n^2) behaviour again.

### DIFF
--- a/lib/src/parser.dart
+++ b/lib/src/parser.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:collection';
+
 import 'arg_parser.dart';
 import 'arg_parser_exception.dart';
 import 'arg_results.dart';
@@ -36,10 +38,18 @@ class Parser {
   /// The accumulated parsed options.
   final Map<String, dynamic> results = <String, dynamic>{};
 
-  Parser(this.commandName, this.grammar, this.args,
-      [this.parent, List<String> rest]) {
+  Parser(this.commandName, this.grammar, List<String> args,
+      [this.parent, List<String> rest])
+      : args = _toReverseList(args) {
     if (rest != null) this.rest.addAll(rest);
   }
+
+  /// Converts a `List` to a [_ReverseList], unless it already is one.
+  ///
+  /// We rely on the 'already is one' case when sharing args lists between
+  /// parsers, so they see the same state.
+  static _ReverseList _toReverseList(List<String> args) =>
+      args is _ReverseList ? args : _ReverseList.of(args);
 
   /// The current argument being parsed.
   String get current => args[0];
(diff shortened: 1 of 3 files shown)


davidmorgan:
Hi Jake,

I noticed that there's appreciable overhead from args parsing in really big build actions. The fix is not terribly pretty, though. WDYT?

Thanks :)

lib/src/parser.dart
```diff
@@ -300,3 +310,47 @@ class Parser {
...
   }
 }
+
+/// A `List` which uses another `List` as its underlying data structure, but
+/// with the elements in reverse.
+///
+/// The makes `removeAt(0)` much faster, which is critical for parser
+/// performance.
+///
+/// It would be better to use a `Queue`, but `List` was already part of the
```
jakemac53: Note that package:collection does have a `QueueList` - likely not worth picking up the dependency though

davidmorgan:
Yeah, I tried that :)

Unfortunately the way it works with "commands" is that it creates a sub parser, and shares the list. It then calls methods on the "parent" that it expects to modify the shared list immediately.

This looks like the least worst change that's not a breaking API change :) ... happy for suggestions of course.

lib/src/parser.dart
davidmorgan: Ah, along with a comment that it's basically a copy from dart:collection :/ ... that's annoying. I don't think I want to create a third copy to avoid the dep :)

lib/src/parser.dart
```diff
@@ -300,3 +310,47 @@ class Parser {
...
+
+  _ReverseList.of(List<String> list) : _list = list.reversed.toList();
+
+  @override
+  int get length => _list.length;
+
+  @override
+  set length(int length) {
+    // Elements are reversed, so if we need to drop elements, drop them from
+    // the end; there's no need to support increasing length.
```
jakemac53: s/end/start

lib/src/parser.dart
```diff
@@ -300,3 +310,47 @@ class Parser {
...
+
+  @override
+  String removeAt(int index) => _list.removeAt(_reversedIndex(index));
+
+  @override
+  String operator [](int index) => _list.elementAt(_reversedIndex(index));
+
+  @override
+  void operator []=(int index, String value) =>
+      _list[_reversedIndex(index)] = value;
```
jakemac53: should we both throwing a custom out of bounds exception (it will give a weird one with a negative number I think today?)

lib/src/parser.dart
```diff
@@ -300,3 +310,47 @@ class Parser {
...
+  @override
+  set length(int length) {
+    // Elements are reversed, so if we need to drop elements, drop them from
+    // the end; there's no need to support increasing length.
+    if (length < _list.length) {
+      _list.removeRange(0, _list.length - length);
+    } else if (length == _list.length) {
+      // Nothing to do.
+    } else {
+      throw UnsupportedError("Can't increase length.");
```
jakemac53: We could support this by inserting nulls at the front... if we are exposing this list to users we should probably make it work as much like a normal list as possible?

davidmorgan:
Thanks, comments look good--done for the day though, will get back to this next week :)

jakemac53:
Sounds good, and nice catch! :rocket:

kevmoo:
Just curious: have we found a place where handling 10k args is a thing? 😄

jakemac53:
> Just curious: have we found a place where handling 10k args is a thing?

At least close to it yes lol. The compilers take all transitive modules as kernel file inputs, and each is listed as a command line argument. So if you have a large app with small modules :boom:

lib/src/parser.dart
davidmorgan: Done.

davidmorgan:
Different approach here, what do you think:

https://github.com/dart-lang/args/pull/130

@kevmoo actually we're easily over 10k in some cases in google3 :) ... note that changing the arg format from `--foo=bar` to `--foo bar` is allowed, and doubles the number of args :)

davidmorgan:
Dropping this one, going with the `Queue` approach. Thanks.

PR #130, opened 2020-02-03
Fix performance issues due to list.removeAt(0), which makes parsing O(n^2).

davidmorgan:
Add a test that checks that parse time does not increase more than 30x when input size grows from 10k to 100k. The current increase is about x2, so this is a very loose check that should be reliably green. It will only trigger if we hit O(n^2) behaviour again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## 1.5.3
 
-* Improve performance for large numbers of args. Parsing 10k args fell from
-  250ms to 50ms; parsing 100k args fell from 15s to 100ms.
+* Improve performance for large numbers of args.
 
 ## 1.5.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.5.3
+
+* Improve performance for large numbers of args. Parsing 10k args fell from
+  250ms to 50ms; parsing 100k args fell from 15s to 100ms.
+
 ## 1.5.2
 
 * Added support for `usageLineLength` in `CommandRunner`
--- a/test/parse_performance_test.dart
+++ b/test/parse_performance_test.dart
@@ -1,0 +1,35 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:args/args.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('ArgParser.parse()', () {
+    test('is fast', () {
+      var parser = ArgParser()..addFlag('flag');
+
+      var baseSize = 10000;
+      var baseList = List<String>.generate(baseSize, (_) => '--flag');
+
+      var multiplier = 10;
+      var largeList =
+          List<String>.generate(baseSize * multiplier, (_) => '--flag');
+      var baseTime = _time(() => parser.parse(baseList));
+      var largeTime = _time(() => parser.parse(largeList));
+      expect(largeTime, lessThan(baseTime * multiplier * 3),
+          reason:
+              'Comparing large data set time ${largeTime}ms to small data set time '
+              '${baseTime}ms. Data set increased ${multiplier}x, time is allowed to '
+              'increase up to ${multiplier * 3}x, but it increased '
+              '${largeTime ~/ baseTime}x.');
+    });
+  });
+}
+
+int _time(void Function() function) {
+  var stopwatch = Stopwatch()..start();
+  function();
+  return stopwatch.elapsedMilliseconds;
+}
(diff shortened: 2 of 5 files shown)


davidmorgan:
This is an alternative to https://github.com/dart-lang/args/pull/128

I realized that `Parser` is internal to the library; you'd have to have a /src/ import to see this API. I checked in google3 and there aren't any, nobody has visibility to this class. So I think this is a pretty safe change, and better/simpler than the other approach with `_ReversedList`.

I timed both, this is actually very slightly slowed than `_ReversedList`. But not in any way that matters, this is still a big speedup for all cases.

```
small (10k)
current/ms: 445, 432, 403, 409, 450, 418
ReversedList/ms: 45, 44, 46, 47, 47, 41, 51, 49, 47
Queue/ms: 42, 36, 45, 40, 46, 38, 50, 67, 42

large (100k)
current/ms: 30k, 30k, 30k, 29k, 29k, 29k, 29k, 29k
ReversedList/ms: 75, 99, 91, 109, 74, 84, 89, 96, 107
Queue/ms: 82, 104, 104, 103, 106, 133, 111, 69, 114
```

What do you think, please?

jakemac53:
> I timed both, this is actually very slightly slowed than `_ReversedList`. But not in any way that matters, this is still a big speedup for all cases.

Likely because you initialize it with a length? Either way it seems fine though.

What I actually find most interesting here is how they both scale sub-linearly? Is there some large cost we are paying for parsing even small #s of args?

davidmorgan:
I guess `Queue` is a bit slower because it's a linked list--which is what you need to allow adds/removes at _either_ end, but since we need only one end, an array-based `List` is slightly better :)

I guess the sublinear scaling means there is some other fixed overhead, maybe JIT compilation; I didn't spend a long time getting the benchmark to be high quality since 30s->0.1s is conclusive already :D 

Thanks.

jakemac53:
> I guess the sublinear scaling means there is some other fixed overhead, maybe JIT compilation; I didn't spend a long time getting the benchmark to be high quality since 30s->0.1s is conclusive already :D

Right I think this is a good change regardless but it does signal that there might be some other easy wins.

To help eliminate noise from the JIT you could use aot mode or you could use the `benchmark_harness` package which does some warm up runs before measuring.

test/parse_performance_test.dart
```diff
@@ -0,0 +1,35 @@
...
+      var parser = ArgParser()..addFlag('flag');
+
+      var baseSize = 10000;
+      var baseList = List<String>.generate(baseSize, (_) => '--flag');
+
+      var multiplier = 10;
+      var largeList =
+          List<String>.generate(baseSize * multiplier, (_) => '--flag');
+      var baseTime = _time(() => parser.parse(baseList));
+      var largeTime = _time(() => parser.parse(largeList));
```
jakemac53: nit: do one of these in `setUp`. This is likely capturing a lot of jit warmup time for the base list parse time.

davidmorgan: Right now this test doesn't print anything--I edited it locally to get the numbers on the review thread. If I improve the benchmark, should I also make it print timings every time?

davidmorgan:
True, it certainly wouldn't hurt to save another 45ms. I'll dig a bit more. I guess, if I find any other improvements I'll split those into different PRs, though.

I take it you're happy with the `Queue` approach?

Thanks :)

jakemac53:
> I take it you're happy with the `Queue` approach?

Yes I think this is better as long as its not a significant regression in terms of performance :).

test/parse_performance_test.dart
jakemac53: You could print - but I was more concerned about the multiplier actually being skewed a lot because of the base time being much higher than it should be :)

davidmorgan: Done--added the print, it seems harmless enough and it's kind of nice to have.

With the warmup the numbers are much closer to 10x, as expected. They're also faster, so I increased the 'base' number so the whole thing takes about a second, to reduce the noise while not annoying people too much.

```
Parsed 200000 elements in 87ms, 2000000 elements in 818ms.
```

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.5.3
+
+* Improve performance for large numbers of args. Parsing 10k args fell from
+  250ms to 50ms; parsing 100k args fell from 15s to 100ms.
```
jakemac53: update these numbers?

davidmorgan: Removed them :) ... it's a bit awkward to get the numbers, and I plan on following up with a second improvement, looks like we can get about another 2x by using string matching instead of regexps. If it works for you (and if you think the numbers are worth having) I'll add them back then? Thanks.

davidmorgan:
(Apologies, now actually pushed the commits).

Looks like `pubspec.yaml` was updated already.

CHANGELOG.md
jakemac53: sgtm - just don't want to have bad numbers if we list them :D

davidmorgan:
Thanks :)

Issue #123, opened 2019-10-24
ArgParser.allowAnything not supported for subcommands


leoafarias:
When a subcommand has an argParser of `ArgParser.allowAnything()` we get the following  error when running as a sub command of a **Command Runner**

```
UnsupportedError (Unsupported operation: ArgParser.allowAnything().addFlag() isn't supported.)
```

I have looked at the tests, and it seems that does not currently account when running the sub command as part of a CommandRunner. I believe the issue is happening on the way the Command argParser is receiving the argParser of the sub commands.

I have gone ahead and done a PR with the tests that it should pass.

leoafarias:
After doing some more research it seems that there still an issue that I am trying to understand.
The following operator https://github.com/dart-lang/args/blob/master/lib/src/arg_results.dart#L62 throws the exception issue when used within the following line:

https://github.com/dart-lang/args/blob/master/lib/command_runner.dart#L180

If I am not mistaken this conditional above and operator is never evaluated, and throws an exception.

PR #124, opened 2019-10-24
Compatibility for allowAnything for Subcommands on CommandRunner

leoafarias:
This pull request is meant to fix issue #123

PR #132, opened 2020-02-21
address an issue when using ArgParser.allowAnything() with CommandRunner

devoncarew:
Address an issue when using `ArgParser.allowAnything()` with `CommandRunner`:
- No longer automatically add a 'help' option to commands that don't validate
  their arguments
- fix https://github.com/dart-lang/args/issues/123 (a slight variant of the PR https://github.com/dart-lang/args/pull/124)

The CommandRunner class currently doesn't work well with Commands that use an ArgParser.allowAnything() parser; this PR addresses that. It'll allow for subcommands that can do things like just delegate all arg parser (and execution) to another entity, like a subprocess.

### DIFF
--- a/lib/command_runner.dart
+++ b/lib/command_runner.dart
@@ -177,7 +177,7 @@ class CommandRunner<T> {
       commands = command._subcommands;
       commandString += ' ${argResults.name}';
 
-      if (argResults['help']) {
+      if (argResults.options.contains('help') && argResults['help']) {
         command.printUsage();
         return null;
       }
@@ -364,8 +364,10 @@ abstract class Command<T> {
   List<String> get aliases => const [];
 
   Command() {
-    argParser.addFlag('help',
-        abbr: 'h', negatable: false, help: 'Print this usage information.');
+    if (!argParser.allowsAnything) {
+      argParser.addFlag('help',
+          abbr: 'h', negatable: false, help: 'Print this usage information.');
+    }
   }
 
   /// Runs this command.
--- a/test/test_utils.dart
+++ b/test/test_utils.dart
@@ -220,7 +220,9 @@ class AllowAnythingCommand extends Command {
   final argParser = ArgParser.allowAnything();
 
   @override
-  Future run() => Future.value().then((_) => hasRun = true);
+  void run() {
+    hasRun = true;
+  }
 }
 
 void throwsIllegalArg(function, {String reason}) {
--- a/test/test_utils.dart
+++ b/test/test_utils.dart
@@ -207,6 +207,22 @@ class AsyncCommand extends Command {
   Future run() => Future.value().then((_) => hasRun = true);
 }
 
+class AllowAnythingCommand extends Command {
+  var hasRun = false;
+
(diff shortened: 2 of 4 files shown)


test/test_utils.dart
```diff
@@ -207,6 +207,22 @@ class AsyncCommand extends Command {
...
+  final name = 'allowAnything';
+
+  @override
+  final description = 'A command using allowAnything.';
+
+  @override
+  final argParser = ArgParser.allowAnything();
+
+  @override
+  Future run() => Future.value().then((_) => hasRun = true);
```
kevmoo: just use `async` here?

lib/command_runner.dart
```diff
@@ -364,8 +364,10 @@ abstract class Command<T> {
   List<String> get aliases => const [];
 
   Command() {
-    argParser.addFlag('help',
-        abbr: 'h', negatable: false, help: 'Print this usage information.');
+    if (!argParser.allowsAnything) {
```
kevmoo: Are we confident this won't break anyone?

test/test_utils.dart
jakemac53: you would need to `await null` before this line if you do that.

lib/command_runner.dart
jakemac53: If I recall the `allowAnything` arg parser never worked with a `Command` anyways. So I think this is likely fine outside of maybe some weird edge cases where somebody implemented `ArgParser`?

jakemac53:
We should ideally test this in flutter and internal before landing though as it is a bit risky

kevmoo:
ack

devoncarew:
Thanks for the review! Agree w/ the testing before a larger rollout.

Wrt flutter, I think this will allow them to improve the way they expose things like pub. Currently w/ the way their command dispatch is implemented, if you run something like `flutter pub upload -h` you don't get any of the `uploader` options printed even thought uploader has ~2 options. Changing to this style delegating would fix that. (cc @jonahwilliams for fyi)

test/test_utils.dart
natebosch: Why are we relying on there being a microtask before setting the value?

devoncarew: This is taken from the command impl. above, but it looks like it may not be necessary.

lib/command_runner.dart
devoncarew: If they were using the `allowAnything` parser, the `argParser.addFlag()` call below would throw an exception (so, I don't think any existing code will be broken here).

test/test_utils.dart
devoncarew: (changed to just `void run() { hasRun = true; }`)

devoncarew:
Landing this; will likely publish as a dev release, and test that through the sdk/flutter/google3 repos.

natebosch:
> will likely publish as a dev release, and test that through the sdk/flutter/google3 repos.

The Dart SDK and google3 don't need a version published in pub.

PR #133, opened 2020-02-26
prep for publishing 1.5.3

devoncarew:
- prep for publishing `1.5.3`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.5.3-dev
+## 1.5.3
 
 * Improve arg parsing performance for large numbers of args.
 * No longer automatically add a 'help' option to commands that don't validate
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.5.3
+## 1.5.3-dev
 
 * Improve arg parsing performance for large numbers of args.
 * No longer automatically add a 'help' option to commands that don't validate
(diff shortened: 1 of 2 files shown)


kevmoo:
Call it `-beta`? Then we distinguish between a release we're working on vs something shipped.

natebosch:
Why do we need to publish as non-stable?

See https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change which explains our use of the `-dev` suffix

devoncarew:
> Why do we need to publish as non-stable?

I'm also happy to publish this as the next stable release; I expect there's little chance that the change will be breaking for existing uses of the API.

natebosch:
Have we written any code in google3 or the SDK or flutter that uses this?

https://github.com/dart-lang/args/pull/132#issuecomment-590475671

natebosch:
cc @jakemac53

devoncarew:
> Have we written any code in google3 or the SDK or flutter that uses this?
> 
> [#132 (comment)](https://github.com/dart-lang/args/pull/132#issuecomment-590475671)
> 
> An unmerged PR with a git override would be great for this.

There are no plans to take advantage of this change in google3 or flutter.

Google3 does pass w/ this change.

The SDK CL is here: https://dart-review.googlesource.com/c/sdk/+/137653. It also passes the presubmit tests.

devoncarew:
fyi, publishing this now

PR #136, opened 2020-03-08
Allow marking commands as deprecated.

### LINKED ISSUE #125: Allow commands to be marked as deprecated
It would be helpful if commands could be marked as deprecated so users know they may soon be removed.  This could be indicated automatically in usage text and output a warning when the command is run.

masterashu:
This pull request fixes #125 .
We can pass an argument to an ArgParser to mark that command as deprecated, which will show warning when used.

### DIFF
--- a/lib/src/parser.dart
+++ b/lib/src/parser.dart
@@ -70,6 +70,10 @@ class Parser {
       if (command != null) {
         validate(rest.isEmpty, 'Cannot specify arguments before a command.');
         var commandName = args.removeFirst();
+        if (command.isDeprecated) {
+          print(command.deprecationMessage);
+        }
+
         var commandParser = Parser(commandName, command, args, this, rest);
 
         try {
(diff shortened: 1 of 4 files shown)


masterashu:
@googlebot I signed it!

lib/src/parser.dart
```diff
@@ -66,6 +66,10 @@ class Parser {
       if (command != null) {
         validate(rest.isEmpty, 'Cannot specify arguments before a command.');
         var commandName = args.removeFirst();
+        if (command.isDeprecated) {
+          print('Deprecation Warning: ${command.deprecationMessage}');
```
natebosch: We should not `print` here, since we don't know that the normal output for `print` is what the author is using to communicate to the user.

natebosch:
Closing since the approach for handling the warning needs some thought.

PR #137, opened 2020-03-13
Remove extra blank lines between options

### LINKED ISSUE #48: Provide an option not to add a newline after multiline option help
Consider:

```
$ flutter drive --help
Runs Flutter Driver tests for the current project.

Usage: flutter drive [arguments]
-h, --help                     Print this usage information.
    --[no-]checked             Toggle Dart's checked mode.
                               (defaults to on)

    --[no-]trace-startup       Start tracing during startup.
    --route                    Which route to load when starting the app.
-t, --target                   Target app path / main entry-point file.
                               (defaults to "lib/main.dart")

    --[no-]keep-app-running    Will keep the Flutter application running when done testing. By default Flutter Driver stops the application after tests are finished.
    --[no-]use-existing-app    Will not start a new Flutter application but connect to an already running instance. This will also cause the driver to keep the application running after tests are done.
    --debug-port               Listen to the given port for a debug connection.
                               (defaults to "8181")

Run "flutter help" to see global options.
```

The newlines after `--checked` and `--target` make it look like the options are grouped into three groups, which they are not.

natebosch:
Fixes #48

The extra blank lines imply a grouping, however the grouping is
arbitrary based on which options happened to wrap, and may vary based on
the terminal width. Instead keep all options without extra newlines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
-## 1.5.3-dev
+## 1.6.0
+
+* Remove the blank lines in usage which separated the help for options that
+  happened to span multiple lines.
+
+## 1.5.3
 
 * Improve arg parsing performance: use queues instead of lists internally to
   get linear instead of quadratic performance, which is important for large
--- a/lib/src/usage.dart
+++ b/lib/src/usage.dart
@@ -112,12 +112,6 @@ class Usage {
           write(2, '(defaults to "${option.defaultsTo}")');
         }
       }
-
-      // If any given option displays more than one line of text on the right
-      // column (i.e. help, default value, allowed options, etc.) then put a
-      // blank line after it. This gives space where it's useful while still
-      // keeping simple one-line options clumped together.
-      if (numHelpLines > 1) newline();
     }
 
     return buffer.toString();
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: args
-version: 1.5.3
+version: 1.6.0-dev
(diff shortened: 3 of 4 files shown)


natebosch:
cc @jonahwilliams

jonahwilliams:
https://github.com/flutter/flutter/pull/52552 for testing this change against the flutter tools presubmit

jonahwilliams:
Analyze check failure is expected due to dependency override

jonahwilliams:
Confusingly with this change, we now have some unexpected test failures unrelated to parsing. (Are there other changes in the args package that haven't been released?)

https://github.com/flutter/flutter/blob/master/packages/flutter_tools/test/commands.shard/hermetic/precache_test.dart#L79

https://cirrus-ci.com/task/6256601757646848

natebosch:
There were some performance optimizations made by @davidmorgan and @jakemac53 - I didn't review them so I don't have a guess as to whether they could be involved in this.

From the failure it looks like the last argument you passed got ignored?

What version were you synced to before so that we can narrow down the changes?

jonahwilliams:
We're on 1.5.2 right now. At any rate, everything else looks good!

davidmorgan:
Certainly could be related, I'll take a look.

davidmorgan:
Yep--the change didn't allow underscores. Sorry about that. Sent https://github.com/dart-lang/args/pull/141

jonahwilliams:
Thanks for the fix @davidmorgan

PR #139, opened 2020-03-13
Reference CommandRunner from addCommand docs

### LINKED ISSUE #134: ArgParser.usage does not print/list out the available commands
When using commands in ArgParser, `usage`  does not print the usage of available commands.
for example
```dart
main(List<String> args){
  var argsParser = ArgParser();
  argsParser.addFlag('flag1', help: 'Help for flag1');
  var argsParser2 = ArgParser();
  argsParser2.addFlag('flag2', help: 'Help for flag2');
  argsParser2.addOption('name', abbr: 'n', help: 'Help for option name');
  argsParser.addCommand('cmd', argsParser2);
  print(argsParser.usage);
}
```
When run the output comes as following:
> --[no-]flag1    Help for flag1  

I believe it would be better to have a description for a command such as
> --[no-]flag1    Help for flag1  
> cmd                Really helpful description for cmd

natebosch:
Closes #134

The `addCommand` API is a little confusing, it doesn't impact the usage
string the way you might expect.

In the future we might consider hiding this API, for now call out the
confusing aspect and point to the API that most folks should use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 1.5.4-dev
 
+* Point towards `CommandRunner` in the docs for `ArgParser.addCommand` since it
+  is what most authors will want to use instead.
+
 ## 1.5.3
 
 * Improve arg parsing performance: use queues instead of lists internally to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-## 1.5.3-dev
+## 1.5.4-dev
+
+## 1.5.3
 
 * Improve arg parsing performance: use queues instead of lists internally to
   get linear instead of quadratic performance, which is important for large
--- a/lib/src/arg_parser.dart
+++ b/lib/src/arg_parser.dart
@@ -8,7 +8,6 @@
 
 import 'dart:collection';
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
-## 1.5.3-dev
+## 1.5.4-dev
```
jakemac53: Probably worth an entry here too

lib/src/arg_parser.dart
```diff
@@ -8,6 +8,7 @@
 
 import 'dart:collection';
 
+import '../command_runner.dart';
```
jakemac53: It might not be worth pulling this in just for the doc comment reference - I think it might actually pull in a library that otherwise wouldn't be imported if you only import the main library.

CHANGELOG.md
natebosch: Done

lib/src/arg_parser.dart
natebosch: Fair enough. Removed and referenced the class with backticks instead. It'll be a little harder to navigate to but I'd hope most users can figure out how to find it.

PR #145, opened 2020-05-28
Add type safety to `addOption` method.

wurikiji:
To use Dart's type system, add type safety to `addOption`. From this patch, `defaultsTo` can be any type of data. 

### Before this patch

```dart
// when we add arg to argParser
argParser.addOption('intOption', defaultsTo: '200');

// , and when we use this arg
int parsedArg = int.parse(argResults['intOption']);
```

### From this patch

```dart
// when we add arg to argParser
argParser.addOption('intOption', defaultsTo: 200);

//, and when we use this arg
int parsedArg = argResults['intOption'];
```


jakemac53:
I noticed some tests are failing (see the travis link), we also should get some new tests added for this.

I think the primary reason for not having this functionality up until now was due to the potential breaking nature of this change. However, we could possibly roll this together with an NNBD release?

@natebosch wdyt?

natebosch:
This won't work. 

> ```dart
> //, and when we use this arg
> int parsedArg = argResults['intOption'];
> ```

If the argument _was_ passed, this line will fail.

natebosch:
I don't think that solving the problem with a mismatched type is possible. There is no way for this code to know how to decode a `String` into a `T` for all possible `T`. We _could_ consider hardcoding support for a few well known types like `int`, but that starts to feel like a quite leaky API since we can' express a constraint that `T` is one of the supported types. Outside of that we could consider requiring a `T Function(String)` passed to do the conversion, but that definitely would be breaking.

I'm not sure how often the `int` problem comes up. I agree that it doesn't feel quite ergonomic to handle the defaulting and `int.parse`. If we decide to do anything here it might be safest to add a new `addIntOption` call - both `String` and `int` would be the supported types for options. That feels a little specialized to me and still doesn't solve the problem of other types.

My preference here would be to not try to solve this, and to keep the `String` to some other type conversion handled in the calling code.

jakemac53:
Does it work to supply any type of object for `default` today? Even if it is possible, I would be surprised if a significant number of users are supplying a default in practice other than trivially supportable types (`bool`, `int`, `String`, `List<int>`, `List<String>`, etc).

Supporting those types by default would mean no static type safety around the unsupported types - you would get a runtime exception when we see T not being one of the supported types and no function provided to do the conversion.

But, it would otherwise be non-breaking for most folks.

natebosch:
I filed https://github.com/dart-lang/args/issues/173 to keep this in mind in the future. I'm going to close this for now since the current approach isn't likely to work. Thanks for the discussion!

PR #148, opened 2020-07-09
improved the doco on the CommandRunner

bsutton:
I've submitted this PR to replace #127 as that PR has languished do to a the lack of a contributor signoff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,16 +270,37 @@ the left of `"commit"`, it is given to the top-level parser.
 If you're writing a command-based application, you can use the [CommandRunner][]
 and [Command][] classes to help structure it. [CommandRunner][] has built-in
 support for dispatching to [Command][]s based on command-line arguments, as well
-as handling `--help` flags and invalid arguments. For example:
+as handling `--help` flags and invalid arguments.
+
+When using the [CommandRunner][] it replaces the [ArgParser][].
+
+In the following example we build a dart application called `dgit` that takes commands `commit` and `stash`.
+
+The [CommandRunner][] takes an 'executableName' which is ONLY used to generate the help message.
+
+> Note: for you final application to be called `dgit` rather than `dgit.dart` you would need to compile it using dart2native.
+
+e.g.
+`dgit commit -a`
+
+File `dgit.dart`
 
 ```dart
-var runner = CommandRunner("git", "Distributed version control.")
-  ..addCommand(CommitCommand())
-  ..addCommand(StashCommand())
-  ..run(['commit', '-a']); // Calls [CommitCommand.run()]
+
+void main(List<String> args)
+{
+  var runner = CommandRunner("mygit", "A dart implementation of distributed version control.")
+    ..addCommand(CommitCommand())
+    ..addCommand(StashCommand())
+    ..run(args); 
 ```
 
-Custom commands are defined by extending the [Command][] class. For example:
+When the above `run(args)` line executes it process the command line args looking for one of the commands (`commit` or `stash`).
+
+If the [CommandRunner][] finds a matching command then the [CommandRunner][] calls the overloaded `run()` method on the matching command (e.g. CommitCommand().run).
+
+
+Commands are defined by extending the [Command][] class. For example:
 
 ```dart
 class CommitCommand extends Command {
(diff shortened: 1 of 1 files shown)


bsutton:
The CI failure here has nothing to do with my changes to the readme. Looks like a bug in the test packages used.

bsutton:
Any chance of getting this pr merged.  
Its non-breaking :)

README.md
```diff
@@ -270,16 +270,39 @@ the left of `"commit"`, it is given to the top-level parser.
...
-as handling `--help` flags and invalid arguments. For example:
+as handling `--help` flags and invalid arguments.
+
+When using the [CommandRunner][] it replaces the [ArgParser][].
+
+In the following example we build a dart application called `dgit` that takes commands `commit` and `stash`.
+
+The [CommandRunner][] takes an 'executableName' which is ONLY used to generate the help message.
+
+> Note: for you final application to be called `dgit` rather than `dgit.dart` you would need to compile it using dart2native.
```
natebosch, line 281: Drop this line. `dart2native` is not the common approach for this. `pub global activate` is much more common - but in either case it doesn't need to be documented here.

README.md
```diff
@@ -270,16 +270,39 @@ the left of `"commit"`, it is given to the top-level parser.
...
 and [Command][] classes to help structure it. [CommandRunner][] has built-in
 support for dispatching to [Command][]s based on command-line arguments, as well
-as handling `--help` flags and invalid arguments. For example:
+as handling `--help` flags and invalid arguments.
+
+When using the [CommandRunner][] it replaces the [ArgParser][].
+
+In the following example we build a dart application called `dgit` that takes commands `commit` and `stash`.
+
+The [CommandRunner][] takes an 'executableName' which is ONLY used to generate the help message.
```
natebosch, line 279: I don't know why we need the emphasis on "only" here - it seem superfluous to me.

```suggestion
The [CommandRunner][] takes an `executableName` which is used to generate the help message.
```

README.md
```diff
@@ -270,16 +270,39 @@ the left of `"commit"`, it is given to the top-level parser.
...
+
+> Note: for you final application to be called `dgit` rather than `dgit.dart` you would need to compile it using dart2native.
+
+e.g.
+`dgit commit -a`
+
+File `dgit.dart`
```
natebosch, line 286: Remove this line, I don't think we need to get into details like this - code sample is the focus here.

README.md
```diff
@@ -270,16 +270,39 @@ the left of `"commit"`, it is given to the top-level parser.
...
 
 ```dart
-var runner = CommandRunner("git", "Distributed version control.")
-  ..addCommand(CommitCommand())
-  ..addCommand(StashCommand())
-  ..run(['commit', '-a']); // Calls [CommitCommand.run()]
+
+void main(List<String> args)
+{
```
natebosch, line 291:
```suggestion
void main(List<String> args) {
```

README.md
```diff
@@ -270,16 +270,39 @@ the left of `"commit"`, it is given to the top-level parser.
...
+    ..addCommand(StashCommand())
+    ..run(args); 
 ```
 
-Custom commands are defined by extending the [Command][] class. For example:
+When the above `run(args)` line executes it parses the command line args looking for one of the commands (`commit` or `stash`).
+
+
+
+If the [CommandRunner][] finds a matching command then the [CommandRunner][] calls the overloaded `run()` method on the matching command (e.g. CommitCommand().run).
```
natebosch, line 302: "overloaded" isn't the right word here. "overridden" would be better, but I think unnecessary.

```suggestion
If the [CommandRunner][] finds a matching command then the [CommandRunner][] calls the `run()` method on the matching command (for example `CommitCommand().run`).
```

README.md
bsutton: I don't really agree with this one.
Adding the command line usage improves comprehension.

During my first read through the doco I missed the code exampling showing (run(['commit', '-a')).

Showing the entire usage path is always useful and certainly does not harm.

README.md
bsutton: removed.

natebosch (approved):

PR #158, opened 2020-11-05
null safety cleanup

jakemac53:
* update sdk constraints and deps
* remove experiment flags from CI
* fix some tests

Note that I had already pushed an update to bump the version to 2.0.0-nullsafety. Feel free to push back on that here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ jobs:
       name: "Analyze (with experiment flag)"
       dart: dev
       os: linux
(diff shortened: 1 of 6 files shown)


natebosch:
Could we merge `null_safety` into `master` first, and then I'll take a look at landing some breaking changes?

I don't think we should land this onto the `null_safety` branch.

jakemac53:
This is just cleaning things up a bit to prepare for the merge into master, I would rather clean it up _before_ merging in to master

natebosch (approved):

PR #159, opened 2020-11-05
Migrate to null safety

jakemac53:
- merges the null safety branch into master
- this isn't ready yet, but further development will happen on the master branch

This migration never received a review afaict, so it needs one at this time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ jobs:
       script: pub run test -p chrome
 
 branches:
-  only: [master, null_safety]
+  only: [master]
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ jobs:
       name: "Analyze (with experiment flag)"
       dart: dev
       os: linux
-      script: dartanalyzer --enable-experiment=non-nullable --fatal-warnings --fatal-infos .
+      script: dartanalyzer --fatal-warnings --fatal-infos .
     - stage: analyze_and_format
       name: "Format"
       dart: dev
@@ -24,12 +24,12 @@ jobs:
       name: "Vm Tests"
       dart: dev
       os: linux
-      script: pub run --enable-experiment=non-nullable test -p vm
+      script: pub run test -p vm
     - stage: test
       name: "Web Tests"
       dart: dev
       os: linux
-      script: pub run --enable-experiment=non-nullable test -p chrome
+      script: pub run test -p chrome
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,11 +26,6 @@ jobs:
       os: linux
       script: pub run test -p chrome
 
-matrix:
-  include:
-  - dart: dev
-    dart_task: dartfmt
-
 branches:
   only: [master, null_safety]
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,43 @@
 language: dart
 
 dart:
-- 2.3.0
 - dev
 
-dart_task:
-- test: -p vm
-- dartanalyzer: --fatal-infos --fatal-warnings .
+jobs:
+  include:
+    - stage: analyze_and_format
+      name: "Analyze lib/ (no experiment flag)"
+      dart: dev
+      os: linux
+      script: dartanalyzer --fatal-warnings --fatal-infos lib/
+    - stage: analyze_and_format
+      name: "Analyze (with experiment flag)"
+      dart: dev
+      os: linux
+      script: dartanalyzer --enable-experiment=non-nullable --fatal-warnings --fatal-infos .
+    - stage: analyze_and_format
+      name: "Format"
+      dart: dev
+      os: linux
+      script: dartfmt -n --set-exit-if-changed .
+    - stage: test
+      name: "Vm Tests"
+      dart: dev
+      os: linux
+      script: pub run --enable-experiment=non-nullable test -p vm
+    - stage: test
+      name: "Web Tests"
+      dart: dev
+      os: linux
+      script: pub run --enable-experiment=non-nullable test -p chrome
 
 matrix:
   include:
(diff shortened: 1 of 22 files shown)


.travis.yml
```diff
@@ -1,21 +1,43 @@
...
-dart_task:
-- test: -p vm
-- dartanalyzer: --fatal-infos --fatal-warnings .
+jobs:
+  include:
+    - stage: analyze_and_format
+      name: "Analyze lib/ (no experiment flag)"
+      dart: dev
+      os: linux
+      script: dartanalyzer --fatal-warnings --fatal-infos lib/
```
natebosch, line 12: I should have noticed this earlier, this step is redundant now with the one below.

.travis.yml
```diff
@@ -1,21 +1,43 @@
...
+    - stage: analyze_and_format
+      name: "Analyze (with experiment flag)"
```
natebosch, line 14:
```suggestion
      name: "Analyze"
```

lib/command_runner.dart
```diff
@@ -108,26 +108,26 @@ class CommandRunner<T> {
   ///
   /// This always returns a [Future] in case the command is asynchronous. The
   /// [Future] will throw a [UsageException] if [args] was invalid.
-  Future<T> run(Iterable<String> args) =>
-      Future.sync(() => runCommand(parse(args)));
+  Future<T?> run(Iterable<String> args) =>
```
natebosch, line 111: I'm worried about this `T?` but I'm not sure if there is much we can do about it. It looks like we rely on being able to return a default for when `--help` is used. We could also consider making a required constructor argument of type `T` for the default value.

test/test_utils.dart
```diff
@@ -232,7 +232,7 @@ void throwsFormat(ArgParser parser, List<String> args) {
 }
 
 Matcher throwsUsageException(message, usage) {
-  return throwsA(predicate((error) {
+  return throwsA(predicate((dynamic error) {
     expect(error, TypeMatcher<UsageException>());
     expect(error.message, message);
     expect(error.usage, usage);
```
natebosch, line 238:
```suggestion
  return throwsA(predicate((Object error) {
    expect(error, isA<UsageException>()
        .having((e) => e.message, 'message', message)
        .having((e) => e.usage, 'usage', usage));
```

lib/command_runner.dart
```diff
@@ -108,26 +108,26 @@ class CommandRunner<T> {
...
 
       var command = commands[error.commands.first];
       for (var commandName in error.commands.skip(1)) {
-        command = command.subcommands[commandName];
+        command = command!.subcommands[commandName];
       }
 
-      command.usageException(error.message);
+      command!.usageException(error.message);
       return null;
```
natebosch, line 131: If we make the return type of `usageException` `Never` can we drop the `return null`?

lib/command_runner.dart
```diff
@@ -142,14 +142,14 @@ class CommandRunner<T> {
   /// here to enable verbose logging before running the command.
   ///
   /// This returns the return value of [Command.run].
-  Future<T> runCommand(ArgResults topLevelResults) async {
-    var argResults = topLevelResults;
+  Future<T?> runCommand(ArgResults topLevelResults) async {
+    ArgResults? argResults = topLevelResults;
```
natebosch, line 146: This seems like we should be able to shift where the `!` is and make the variable non-nullable.

lib/command_runner.dart
```diff
@@ -374,7 +374,7 @@ abstract class Command<T> {
   ///
   /// The return value is wrapped in a `Future` if necessary and returned by
   /// [CommandRunner.runCommand].
-  FutureOr<T> run() {
+  FutureOr<T>? run() {
```
natebosch, line 377: Why does this need to be nullable?

lib/command_runner.dart
natebosch: https://github.com/dart-lang/args/pull/160

.travis.yml
natebosch: #161

test/test_utils.dart
natebosch: We can do even better. https://github.com/dart-lang/args/pull/162

lib/command_runner.dart
natebosch: https://github.com/dart-lang/args/pull/163

lib/command_runner.dart
natebosch: This is also related to the design of how help is handled. I'm not super happy with the design but I don't see any easy wins for improving it. I'll likely just make a couple small tweaks.

natebosch: I'm having a hard time deciding between `FutureOr<T?>` and `FutureOr<T>?` here. The former would allow for a `Future<T?>` while the latter wont. Allowing that feels more consistent, but also unnecessary if we consider that the null return for the help command is the main use case. Do you have any opinion on that @jakemac53 ?

lib/command_runner.dart
jakemac53: Ya I don't think we can really do much here, I wouldn't want to force some arbitrary return value - this would end up requiring a lot of people to pass an explicit `null` for that I think since they don't even use these return values.

jakemac53:
I took another look at the return values of `run` and `runCommand` and I think it really doesn't make sense to try and fix those. 

I think if we required some default return value in the constructor most users would end up passing explicit nulls because these return values aren't even commonly used.

lib/command_runner.dart
jakemac53: I think that `FutureOr<T>?` is better.

The help command is a bit special because it isn't implemented by the user and we just can't really fix it without gross workarounds.

For user commands we should force them to return a valid value.

natebosch:
Yeah, I like that thinking.

natebosch (approved):

PR #168, opened 2020-12-09
Give a specific function type for callback

### LINKED ISSUE #149: Enforce that `callback` function passed to `addOption` has a single argument
The static type is currently `Function` which does not offer help on things like argument count. We can't currently give a complete argument because depending on wither it's a multi option it may be passed either a `String` or `List<String>`, but we can at least type it as `void Function(Never)` to statically ensure it allows a single argument.

natebosch:
Closes #149

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   * Instead of `Option.defaultValue`, use `Option.defaultsTo`.
   * Instead of `OptionType.FLAG/SINGLE/MULTIPLE`, use
     `OptionType.flag/single/multiple`.
+* Add a more specific function type to the `callback` argument of `addOption`.
 
 ## 1.6.0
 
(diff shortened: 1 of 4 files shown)


jakemac53 (approved):
This isn't breaking right? Or at least not for anything that wouldn't have already been broken (could move a runtime error to a static one?).

natebosch:
> This isn't breaking right? Or at least not for anything that wouldn't have already been broken (could move a runtime error to a static one?).

This _would_ have been breaking if we hadn't already dropped the `allowMulti` argument in https://github.com/dart-lang/args/pull/165 - if that still existed it would need to be valid to pass a `void Function(String?)` or `void Function(List<String>?)`. For normal usage you are correct that this turns a runtime error into a static one.

There is still a small edge where this may be breaking. If you tear off the method and need it to be exactly the old signature you will be broken. I can't imagine someone is actually doing that.

jakemac53:
Ya I am not worried about the tearoff case, this LGTM

PR #169, opened 2020-12-11
Refactor usage.dart

natebosch:
- Add a top level method rather than require a constructor call and call
  to `generate()` when there is otherwise no reason to hold on to the
  `Usage` object. This makes it more clear that the implementation is
  following a method object pattern.
- Update doc comment which had drifted from the original implementation
  and no longer referred to the correct arguments.
- Make the `StringBuffer` field final since it is now a guaranteed
  single use object.
- Make everything possible private. This revealed that there was a field
  being updated but never read and can be removed.
- Remove `get` from method names.
- Make the `_columnWidths` late final rather than initialize it manually
  in the call to `generate()`.
- Pull out sub-methods `_writeSeparator` and `_writeOption` from
  `generate` to reduce nesting and the length of the overall method.
- use `.take(size)` instead of `.sublist(size)` since we only need an
  Iterable.
- Use `.first` and `.last` over manual indexing.
- Use `.removeAt(0)` over `.removeRange(0,1)`.
- Rewrite a for loop with `.addAll` calls to a List literal with a
  for loop element.

### DIFF
--- a/lib/src/usage.dart
+++ b/lib/src/usage.dart
@@ -17,6 +17,11 @@ import 'utils.dart';
 ///
 /// It builds the usage text up one column at a time and handles padding with
 /// spaces and wrapping to the next line to keep the cells correctly lined up.
+///
+/// [lineLength] specifies the horizontal character position at which the help
+/// text is wrapped. Help that extends past this column will be wrapped at the
+/// nearest whitespace (or truncated if there is no available whitespace). If
+/// `null` there will not be any wrapping.
 String generateUsage(List optionsAndSeparators, {int? lineLength}) =>
     _Usage(optionsAndSeparators, lineLength).generate();
 
@@ -37,7 +42,7 @@ class _Usage {
   int _currentColumn = 0;
 
   /// The width in characters of each column.
-  late final _columnWidths = _calculateLineWidths();
+  late final _columnWidths = _calculateColumnWidths();
 
   /// How many newlines need to be rendered before the next bit of text can be
   /// written.
@@ -85,7 +90,7 @@ class _Usage {
     if (option.help != null) _write(2, option.help!);
 
     if (option.allowedHelp != null) {
-      var allowedNames = option.allowedHelp!.keys.toList(growable: false);
+      var allowedNames = option.allowedHelp!.keys.toList();
       allowedNames.sort();
       _newline();
       for (var name in allowedNames) {
@@ -101,16 +106,12 @@ class _Usage {
       }
     } else if (option.isMultiple) {
       if (option.defaultsTo != null && option.defaultsTo.isNotEmpty) {
-        _write(
-            2,
-            '(defaults to ' +
-                option.defaultsTo.map((value) => '"$value"').join(', ') +
-                ')');
+        var defaults =
+            (option.defaultsTo as List).map((value) => '"$value"').join(', ');
+        _write(2, '(defaults to $defaults)');
       }
-    } else {
-      if (option.defaultsTo != null) {
-        _write(2, '(defaults to "${option.defaultsTo}")');
-      }
+    } else if (option.defaultsTo != null) {
+      _write(2, '(defaults to "${option.defaultsTo}")');
     }
   }
 
@@ -137,7 +138,7 @@ class _Usage {
     return '      [$allowed]' + (isDefault ? ' (default)' : '');
   }
 
-  List<int> _calculateLineWidths() {
+  List<int> _calculateColumnWidths() {
     var abbr = 0;
     var title = 0;
     for (var option in _optionsAndSeparators) {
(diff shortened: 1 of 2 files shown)


natebosch:
I was searching for any low hanging fruit in the feature requests and while I was exploring the feasibility of one of them was having a hard time understanding this code, so I decided to clean it up.

lib/src/usage.dart
```diff
@@ -18,109 +17,107 @@ import 'utils.dart';
 ///
 /// It builds the usage text up one column at a time and handles padding with
```
parlough, line 18: Somewhere in this doc can you include a reference to [lineLength] with an explanation of its functionality, including what it means for it to be null? Now that we don't have the documentation from the Usage field it may not be completely clear.

lib/src/usage.dart
```diff
@@ -18,109 +17,107 @@ import 'utils.dart';
...
-          write(2, '(defaults to "${option.defaultsTo}")');
-        }
+  void _writeOption(Option option) {
+    _write(0, _abbreviation(option));
+    _write(1, _longOption(option));
+
+    if (option.help != null) _write(2, option.help!);
+
+    if (option.allowedHelp != null) {
+      var allowedNames = option.allowedHelp!.keys.toList(growable: false);
```
munificent, line 88: Is the `growable: false` really worth doing here?

lib/src/usage.dart
```diff
@@ -18,109 +17,107 @@ import 'utils.dart';
...
+      if (option.defaultsTo == true) {
+        _write(2, '(defaults to on)');
+      }
+    } else if (option.isMultiple) {
+      if (option.defaultsTo != null && option.defaultsTo.isNotEmpty) {
+        _write(
+            2,
+            '(defaults to ' +
+                option.defaultsTo.map((value) => '"$value"').join(', ') +
+                ')');
```
munificent, line 108: The wrapping is pretty gnarly here. How about hoisting the `'(defaults to...` parameter to a local variable?

lib/src/usage.dart
```diff
@@ -18,109 +17,107 @@ import 'utils.dart';
...
+      }
+    } else {
```
munificent, line 110: Remove the outer block and just do `else if (...) { ... }`

lib/src/usage.dart
```diff
@@ -133,118 +130,103 @@ class Usage {
...
-  String getAllowedTitle(Option option, String allowed) {
+  String _allowedTitle(Option option, String allowed) {
     var isDefault = option.defaultsTo is List
         ? option.defaultsTo.contains(allowed)
         : option.defaultsTo == allowed;
     return '      [$allowed]' + (isDefault ? ' (default)' : '');
   }
 
-  void calculateColumnWidths() {
+  List<int> _calculateLineWidths() {
```
munificent, line 140: I think `_calculateColumnWidths` is a better name.

lib/src/usage.dart
```diff
@@ -133,118 +130,103 @@ class Usage {
...
-            .addAll(wrapTextAsLines(line, start: start, length: lineLength));
-      }
-
-      lines = wrappedLines;
+    if (column == _columnWidths.length && lineLength != null) {
+      var start =
+          _columnWidths.take(column).reduce((start, width) => start + width);
+      lines = [
+        for (var line in lines)
+          ...wrapTextAsLines(line, start: start, length: lineLength),
```
munificent, line 179: 😍

lib/src/usage.dart
```diff
@@ -133,118 +130,103 @@ class Usage {
...
+      ];
     }
 
     // Strip leading and trailing empty lines.
-    while (lines.isNotEmpty && lines[0].trim() == '') {
-      lines.removeRange(0, 1);
```
munificent, line 194: Haha, wow, This is *old* Dart code. (Dart used to not have `removeAt()` because someone on the team was morally opposed to allowing users the possibility of accidentally calling that in a loop instead of the more efficient `removeRange()`.)

munificent (approved):
A bunch of little comments, but this looks great!

lib/src/usage.dart
natebosch: Done.

lib/src/usage.dart
natebosch: 🤷  it was there before and I didn't want to change it if it _was_ worth doing.

I took a look at it appears it was only added as a precaution when the default changed - it used to be non-growable by default.

https://codereview.chromium.org//12401002


I'll go ahead and remove since it's unlikely to matter and it's noise. (I'm also tempted to use a cascade for the `sort()` call but I guess I'll hold off on that 😝

lib/src/usage.dart
natebosch: Done.

lib/src/usage.dart
natebosch: Done.

lib/src/usage.dart
natebosch: Oops, I mistyped when I did the rename to make this private. Done.

PR #170, opened 2020-12-11
Add a test for passing allowedHelp without allowed

### LINKED ISSUE #25: args: Merge allowedHelp into allowed
<a href="https://github.com/seaneagan"></img></a> **Issue by [seaneagan](https://github.com/seaneagan)**
_Originally opened as dart-lang/sdk#15705_

----

Currently it looks something like:

&nbsp;&nbsp;var allowed = {
&nbsp;&nbsp;&nbsp;&nbsp;'x': 'x help',
&nbsp;&nbsp;&nbsp;&nbsp;'y': 'y help',
&nbsp;&nbsp;&nbsp;&nbsp;'z': 'z help',
&nbsp;&nbsp;}

&nbsp;&nbsp;parser.addOption('foo', allowed: allowed.keys.toList(), allowedHelp: allowed);

Nicer would be:

&nbsp;&nbsp;parser.addOption('foo', allowed: {
&nbsp;&nbsp;&nbsp;&nbsp;'x': 'x help',
&nbsp;&nbsp;&nbsp;&nbsp;'y': 'y help',
&nbsp;&nbsp;&nbsp;&nbsp;'z': 'z help',
&nbsp;&nbsp;});

So allowed: would allow Iterable&lt;String&gt; or Map&lt;String, String&gt;

natebosch:
Adds a test against a potential breaking solution for #25 

There is an existing behavior where `allowedHelp` can influence
the usage text without limiting the accepted inputs. Add a test for
this case specifically, even though it specifically retains the
confusing behavior to avoid breaking it since we don't know who
might be relying on it.

### DIFF
--- a/lib/src/arg_parser.dart
+++ b/lib/src/arg_parser.dart
@@ -161,7 +161,9 @@ class ArgParser {
   /// be included in [usage].
   ///
   /// The [allowedHelp] argument is a map from values in [allowed] to
-  /// documentation for those values that will be included in [usage].
+  /// documentation for those values that will be included in [usage]. If this
+  /// option is provided and [allowed] is `null` the allowed options will be the
+  /// map keys.
   ///
   /// The [defaultsTo] argument indicates the value this option will have if the
   /// user doesn't explicitly pass it in (or `null` by default).
@@ -210,7 +212,9 @@ class ArgParser {
   /// be included in [usage].
   ///
   /// The [allowedHelp] argument is a map from values in [allowed] to
-  /// documentation for those values that will be included in [usage].
+  /// documentation for those values that will be included in [usage]. If this
+  /// option is provided and [allowed] is `null` the allowed options will be the
+  /// map keys.
   ///
   /// The [defaultsTo] argument indicates the values this option will have if
   /// the user doesn't explicitly pass it in (or `[]` by default).
--- a/lib/src/arg_parser.dart
+++ b/lib/src/arg_parser.dart
@@ -161,9 +161,7 @@ class ArgParser {
   /// be included in [usage].
   ///
   /// The [allowedHelp] argument is a map from values in [allowed] to
-  /// documentation for those values that will be included in [usage]. If this
-  /// option is provided and [allowed] is `null` the allowed options will be the
-  /// map keys.
+  /// documentation for those values that will be included in [usage].
   ///
   /// The [defaultsTo] argument indicates the value this option will have if the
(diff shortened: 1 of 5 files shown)


test/parse_test.dart
```diff
@@ -335,6 +335,25 @@ void main() {
...
+      test('does not throw if value is allowed by allowedHelp', () {
+        var parser = ArgParser();
+        parser.addOption('mode',
+            abbr: 'm',
+            allowedHelp: {'debug': 'debug mode', 'release': 'release mode'});
+
+        // does not throw
+        parser.parse(['-mdebug']);
+      });
+
```
munificent, line 356: Would also be good to test that passing both `allowed` and `allowedHelp` treats the former as the canonical list. So if `allowed` is `['a', 'b', 'c']` and `allowedHelp` is `{'a': 'A.', 'b': 'B.'}`, then you can still pass `c`.

lib/src/arg_parser.dart
```diff
@@ -210,7 +212,9 @@ class ArgParser {
   /// be included in [usage].
   ///
   /// The [allowedHelp] argument is a map from values in [allowed] to
-  /// documentation for those values that will be included in [usage].
+  /// documentation for those values that will be included in [usage]. If this
+  /// option is provided and [allowed] is `null` the allowed options will be the
+  /// map keys.
```
munificent, line 217: Does this mean there's no way to show help options for some allowed values while still allowing an open-ended set? That seems like an unfortunate restriction. For example, test runner (which doesn't actually use args, but just as an example) only shows a subset of the most common values for some options but allows a larger set.

munificent (commented):
I'm not sure I'm sold on losing the ability to show help for some values while still allowing an open-ended set. Do I understand this PR right?

lib/src/arg_parser.dart
natebosch: > Does this mean there's no way to show help options for some allowed values while still allowing an open-ended set?

Hmm, yeah it does. That wasn't a tested scenario.

Maybe we should just punt on this for now. It does still seem like a usability bug to need to provide both arguments to get the typical desired behavior, but I'm also not eager to introduce a third argument to support the use case of specific option docs + open ended input.

natebosch: I updated this PR to revert my change and add a test for this behavior. I likely won't bother to pursue this further than that.

PTAL.

munificent: LGTM!

munificent (approved):

Issue #171, opened 2020-12-11
Rename Option.getOrDefault to valueOrDefault, and ArgParser.getDefault to default


natebosch:
I wish we had caught this earlier, normally I would like to have a period of time with a `@deprecated` annotation on the old one to let folks migrate to the new one.

I think we can probably do that here too and we'll just be stuck with the old aliases until the next major version.

I'm _almost_ tempted to do the breaking change, but I'd rather not prevent packages from having a wide constraint across the major version bump and I don't think we have the time to roll out a non-breaking version of the package that adds the deprecations.

cc @munificent

PR #172, opened 2020-12-11
Deprecate methods name get*

natebosch:
Towards #171

### DIFF
--- a/lib/src/arg_parser.dart
+++ b/lib/src/arg_parser.dart
@@ -316,7 +316,7 @@ class ArgParser {
     return value.defaultsTo;
   }
 
-  @Deprecated('use defaultFor instead')
+  @Deprecated('Use defaultFor instead.')
   dynamic getDefault(String option) => defaultFor(option);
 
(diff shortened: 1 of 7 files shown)


parlough (approved):

lib/src/arg_parser.dart
```diff
@@ -307,16 +307,18 @@ class ArgParser {
...
+  /// Returns the default value for [option].
+  dynamic defaultFor(String option) {
     var value = options[option];
     if (value == null) {
       throw ArgumentError('No option named $option');
     }
     return value.defaultsTo;
   }
 
+  @Deprecated('use defaultFor instead')
```
munificent, line 319: Do we tend to format these like proper sentences? `'Use defaultFor instead.'`? I don't think I've ever actually used `@deprecated` before (because all of my APIs are so perfect they never need to be deprecated, of course).

munificent (approved):

lib/src/arg_parser.dart
natebosch: I don' think I've ever thought about it. I'll go ahead and format like sentences.

PR #176, opened 2020-12-17
Migrate to GitHub Actions

athomas:


.github/workflows/test-package.yml
```diff
@@ -0,0 +1,57 @@
...
+      - uses: dart-lang/setup-dart@v0.1
+        with:
+          channel: ${{ matrix.sdk }}
+      - name: Install dependencies
+        run: dart pub get
+      - name: Check formatting
+        run: dart format --output=none --set-exit-if-changed .
+      - name: Analyze code
+        run: dart analyze
+        if: ${{ always() && steps['Install dependencies'].outcome == 'success' }}
```
jakemac53, line 32: You need to add an ID to the `Install dependencies` step and then do `steps.<id>.conclusion == 'success'` for this to work afaik :) (at least this is how we do it in package:mono_repo)

jakemac53: You can also remove the `always() &&` afaik

athomas: Ah thanks!

athomas: Hmm, not sure about removing the `always()`.

> If your if expression does not contain any of the status functions it will automatically result with success().
> https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions

I read that as meaning that if I don't use a status function at all, it will be aborted if the previous one fails because it will assume success(). But maybe the if expression overrides that as well 🤷

jakemac53: As far as I could tell when trying to figure this out for mono_repo, any `if` just completely overrides the default behavior (which is `success()`)

.github/workflows/test-package.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          channel: ${{ matrix.sdk }}
+      - id: install
+        name: Install dependencies
+        run: dart pub get
+      - name: Check formatting
+        run: dart format --output=none --set-exit-if-changed .
+      - name: Analyze code
+        run: dart analyze
+        if: ${{ always() && steps.install.outcome == 'success' }}
```
jakemac53, line 33: It doesn't really matter but also fwiw you don't need the `${{ }}` in `if`, its implied

.github/workflows/test-package.yml
athomas: Ok, I'll try removing it and then I'll upload a commit with a formatting error.

athomas: We need the `always()` or the step won't run (discussed offline).

.github/workflows/test-package.yml
athomas: Done.

.github/workflows/test-package.yml
```diff
@@ -0,0 +1,58 @@
...
+      - uses: actions/checkout@v2
+      - uses: dart-lang/setup-dart@v0.1
+        with:
+          channel: ${{ matrix.sdk }}
+      - name: Install dependencies
+        run: dart pub get
+      - name: Run VM tests
+        run: dart test --platform vm
+      - name: Run Chrome tests
+        run: dart test --platform chrome
```
kevmoo, line 58: don't you want that if magic stuff down here, too?

athomas: I'll update this to match your template changes in dart-lang/.github#7 when that is merged.

kevmoo: 👍

athomas: Done.

kevmoo (approved):

Issue #156, opened 2020-11-01
Allow making options required


AndiDog:
Python's argparse module supports [required=True](https://docs.python.org/3/library/argparse.html#required) to make `-a/--a-long-option` required. This is incredibly helpful if the CLI takes many arguments, which you of course don't want to treat as positional arguments (and require the user to know their order), where some are mandatory.

Admittedly, still calling it "option" is somewhat funky... but it's still a very valid use case.

Suggestion: `addOption('my-long-opt', required: true)`, and `required: true` and `defaultsTo != null` mutually exclusive.

Without this feature, every required argument must be checked with `if (!argResults.wasParsed('my-long-opt')) throw ...`.

PoloLacoste:
Working on it !

I will use `mandatory` instead of `required` because [`required`](https://api.flutter.dev/flutter/meta/required-constant.html) is a top level constant in Dart.

PoloLacoste:
#177

PR #177, opened 2021-01-05
Added mandatory (or required) option

PoloLacoste:
Simple usage :

```dart
var parser = ArgParser();
parser.addOption('flag', mandatory: true);
```

If the parser parse arguments without this option it will throw a `FormatException`.

Closes https://github.com/dart-lang/args/issues/156


PoloLacoste:
@googlebot I signed it!

grouma:
Thanks for the PR!

Could you add a `CHANGELOG` entry and update the version in the pubspec?

lib/src/arg_parser.dart
```diff
@@ -280,9 +282,16 @@ class ArgParser {
       }
     }
 
+    // Make sure the option is not mandatory with a default value
+    if(mandatory && defaultsTo != null) {
+      throw ArgumentError(
+        'The option $name canno\'t be mandatory and have a default value.');
```
grouma, line 288: Typo, should either be `can't` or `cannot`

lib/src/parser.dart
```diff
@@ -92,10 +92,19 @@ class Parser {
       rest.add(args.removeFirst());
     }
 
-    // Invoke the callbacks.
+    // Check if mandatory and invoke existing callbacks.
     grammar.options.forEach((name, option) {
+      var parsedOption = results[name];
+
+      // Check if an option was mandatory and exist
+      // if not throw an exception
```
grouma, line 100: [nit] Missing trailing period.

lib/src/arg_parser.dart
natebosch:
```suggestion
        'The option $name cannot be mandatory and have a default value.');
```

lib/src/option.dart
```diff
@@ -70,6 +72,9 @@ class Option {
   /// addition to `--option a --option b`.
   final bool splitCommas;
 
+  /// Whether this option should be required from usage
```
natebosch, line 75:
```suggestion
  /// Whether this option must be provided for correct usage.
```

README.md
```diff
@@ -74,6 +74,12 @@ The callbacks for all options are called whenever a set of arguments is parsed.
 If an option isn't provided in the args, its callback is passed the default
 value, or `null` if no default value is set.
 
+Options can be mandatory (required) with the flag `mandatory` set to `true`.
```
natebosch, line 77: These semantics match well with `required` from the language. I think we should use that term everywhere rather than the term "mandatory" which gets clarified as "required".


```suggestion
If an option is `required` but not provided, the parser throws an [`ArgParserException`][ArgParserException].
```

README.md
```diff
@@ -74,6 +74,12 @@ The callbacks for all options are called whenever a set of arguments is parsed.
...
+
+```dart
+parser.addOption('mode', mandatory: true);
```
natebosch, line 80:
```suggestion
parser.addOption('mode', required: true);
```

example/test_runner.dart
```diff
@@ -164,6 +164,7 @@ is 'dart file.dart' and you specify special command
   parser.addOption('dart', help: 'Path to dart executable');
   parser.addOption('drt', help: 'Path to content shell executable');
   parser.addOption('dartium', help: 'Path to Dartium Chrome executable');
+  parser.addOption('mandatory', help: 'A mandatory option', mandatory: true);
```
natebosch, line 167:
```suggestion
  parser.addOption('required', help: 'A required option', required: true);
```

lib/src/parser.dart
```diff
@@ -92,10 +92,19 @@ class Parser {
...
+      if (option.mandatory && parsedOption == null) {
+        throw ArgParserException('Option $name is mandatory', [name]);
```
natebosch, line 102:
```suggestion
        throw ArgParserException('Option $name is required.', [name]);
```

lib/src/usage.dart
```diff
@@ -122,6 +122,8 @@ class _Usage {
     var result;
     if (option.negatable!) {
       result = '--[no-]${option.name}';
+    } else if (option.mandatory) {
+      result = '--${option.name}';
```
natebosch, line 126: I don't understand what this condition is accomplishing. Is this supposed to be different from the block below?

lib/src/usage.dart
```diff
@@ -131,6 +133,10 @@ class _Usage {
     return result;
   }
 
+  String _mandatoryOption(Option option) {
+    return option.mandatory ? ' (mandatory)' : '';
```
natebosch, line 137:
```suggestion
    return option.required ? ' (required)' : '';
```

lib/src/usage.dart
PoloLacoste: This is a mistake, sorry !

lib/src/parser.dart
PoloLacoste: Fixed !

lib/src/arg_parser.dart
```diff
@@ -280,9 +282,16 @@ class ArgParser {
       }
     }
 
+    // Make sure the option is not required with a default value
```
grouma, line 285: [nit] Trailing period.

grouma (approved):

lib/src/arg_parser.dart
PoloLacoste: Fixed !

natebosch:
Sorry for the churn here. After some internal discussion we think that `mandatory` is indeed a better name to avoid confusion with the keyword. I've handled the rename back and the merge conflicts from other recent changes.

lib/src/arg_parser.dart
```diff
@@ -299,10 +301,17 @@ class ArgParser {
       }
     }
 
+    // Make sure the option is not mandatory with a default value
```
grouma, line 304: [nit] period.

grouma (approved):

lib/src/arg_parser.dart
natebosch: Done.

jakemac53 (approved):

Issue #181, opened 2021-02-12
Alias names for arguments


Hixie:
It would be really useful if you could specify aliases for argument names. For example, if someone were to create an application with a "--referer" option, it would be nice to be able to rename it to "--referrer" without breaking backwards compatibility with old scripts that use the old name; allowing them to specify an alias would mean that we could use the correct spelling in all the help docs while supporting both the new name and the old name.

jakemac53:
Would we want to support alias abbreviations as well as full names?

Hixie:
For my use case I don't need that, but I have no objection in principle.

Hixie:
(My use case is renaming a bunch of arguments to the `flutter` tool which I discovered were problematic when doing a recent clenup.)

jakemac53:
Note that `ArgParser` exposes a `final Map<String, Option> options;` public field. The obvious way to implement this is to add extra entries to that for each alias, but that could be breaking if we now have duplicate options in that map, depending on how consumers are using the map.

We could just not expose it that way, but it would likely be useful.

Hixie:
I don't suppose in the meantime I can hack this by manually adding entries to that map, can I?

jakemac53:
It is an unmodifiable map view around the real map, so no unfortunately haha.

jakemac53:
This sounds to me in general like a reasonable and useful feature, so I do think its something we can do. But as always it isn't _quite_ as easy as it sounds on the surface haha. I think really though we just need to answer the question of whether all the aliases should be added to that _public_ options map or not.

jakemac53:
@natebosch are you aware of any _implementations_ of `ArgParser` out there? I didn't even consider that but then I noticed this package has one (`AllowAnythingParser`), so I wonder if there might be others...

jakemac53:
Actually I just checked my local pub copy and internal code and didn't find any implementations so we should be good to add new parameters to methods.

munificent:
Aliases sound like a reasonable feature to me.

>  I think really though we just need to answer the question of whether all the aliases should be added to that public options map or not.

If it were, I wouldn't. Aliases exist to be more permissive in what *input* is allowed. Once the ArgParser has consumed the input, it should be normalized to something simple and consistent for the rest of code to consume. For example, we don't put options in that map for their abbreviations.

natebosch:
[...]

PR #182, opened 2021-02-17
Add support for argument name aliases

jakemac53:
Closes https://github.com/dart-lang/args/issues/181

Add `aliases` named argument to `addFlag`, `addOption`, and `addMultiOption`, as well as a public `findByNameOrAlias` method on `ArgParser`. This allows you to provide aliases for an argument name, which eases the transition from one argument name to another.

Keys are _not_ added to the `options` map for each alias.

cc @Hixie

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 2.1.0
+
+* Add `aliases` named argument to `addFlag`, `addOption`, and `addMultiOption`,
+  as well as a public `findByNameOrAlias` method on `ArgParser`. This allows
+  you to provide aliases for an argument name, which eases the transition from
+  one argument name to another.
+
 ## 2.0.0
 
 * Stable null safety release.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.1.0
+## 2.1.0-dev
 
 * Add `aliases` named argument to `addFlag`, `addOption`, and `addMultiOption`,
   as well as a public `findByNameOrAlias` method on `ArgParser`. This allows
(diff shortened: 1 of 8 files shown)


lib/src/arg_parser.dart
```diff
@@ -327,4 +352,8 @@ class ArgParser {
     }
     return null;
   }
+
+  /// Finds the option whose name or alias matches [name], or `null` if no
+  /// option has that name or alias.
+  Option? findByNameOrAlias(String name) => options[_aliases[name] ?? name];
```
natebosch, line 358: Does this need to be public?

jakemac53: That is the simplest way, yes. We do also already have a public `findByAbbreviation` method.

We need to do similar logic inside the `parser.dart` library, which would otherwise be duplicated.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 2.1.0
```
natebosch, line 1:
```suggestion
## 2.1.0-dev
```

I'd like to make sure we get this landed internally before publishing, and we've got another PR to merge and validate. Can we keep this as a `-dev` for now?

natebosch (approved):

test/args_test.dart
```diff
@@ -184,6 +184,40 @@ void main() {
     });
   });
 
+  group('ArgParser.findByNameOrAlias', () {
```
natebosch, line 187: Can we add a test that validates aliases are used in `ArgParser.parse()`?

Hixie:
This is fantastic, thanks so much!!!

CHANGELOG.md
jakemac53: done

test/args_test.dart
jakemac53: done

jakemac53:
PR updating the SDK  here https://dart-review.googlesource.com/c/sdk/+/185486 and I have an internal global presubmit running in advance of that now

PR #183, opened 2021-03-11
Add --undefok support

natebosch:
Some other arg parsing libraries support an `--undefok` option which
suppresses the error that would normally surface if an unknown flag or
option is passed.
For example: https://gflags.github.io/gflags/#special

Add built in support, unconditionally, for `--undefok` in the parser.
The `undefok` option always supports multiple values and comma separated
values.

If an option named `undefok` is added to the `ArgParser` it will
override this behavior. It isn't possible to implement this behavior in
the calling code because the `callback` for when an option is parsed is
not triggered until after all arguments are parsed. Allowing the calling
code to override the `undefok` option, even though it can't implement
the expected semantics, retains backwards compatibility in the unlikely
case there are existing uses of that name.

- Add an `_undefok` field to `Parser` to track the names that suppressed
  as options are parsed. Only the top level parser in a `parent` chain
  will handle `undefok`.
- When parsing an otherwise unknown option named `undefok`, parse the
  value and store in the allowed `_undefok` set.
- When parsing an otherwise unknown option with any other name already
  in `_undefok`, silently drop it, along with any `=value` portion.

This implementation has the following limitations:
- The `--undefok` option must be passed _before_ the unknown option. If
  the arguments come in the other order it will still fail.
- The unknown option must either be a flag, or must use the
  `--name=value` syntax. If it is passed as `--name value` then the
  value will show in `argResults.rest`. This could cause the argument to
  get misinterpreted by the app.

### DIFF
--- a/lib/src/parser.dart
+++ b/lib/src/parser.dart
@@ -34,6 +34,8 @@ class Parser {
   /// The accumulated parsed options.
   final Map<String, dynamic> results = <String, dynamic>{};
 
+  final _undefok = <String>{};
+
   Parser(this.commandName, this.grammar, this.args,
       [this.parent, List<String>? rest]) {
     if (rest != null) this.rest.addAll(rest);
@@ -66,7 +68,8 @@ class Parser {
       if (command != null) {
         validate(rest.isEmpty, 'Cannot specify arguments before a command.');
         var commandName = args.removeFirst();
-        var commandParser = Parser(commandName, command, args, this, rest);
+        var commandParser =
+            Parser(commandName, command, args, this, rest);
 
         try {
           commandResults = commandParser.parse();
@@ -284,8 +287,24 @@ class Parser {
       setFlag(results, option, false);
     } else {
       // Walk up to the parent command if possible.
-      validate(parent != null, 'Could not find an option named "$name".');
-      return parent!.parseLongOption();
+      if (parent != null) {
+        return parent!.parseLongOption();
+      }
+      if (name == 'undefok') {
+        args.removeFirst();
+        if (value == null) {
+          validate(args.isNotEmpty, 'Missing argument for "--undefok".');
+          value = current;
+          args.removeFirst();
+        }
+        _undefok.addAll(value.split(','));
+        return true;
+      }
+      if (_undefok.contains(name)) {
+        args.removeFirst();
+        return true;
+      }
+      validate(false, 'Could not find an option named "$name".');
     }
 
     return true;


natebosch:
I'll need to add some tests if we decide to go forward with this. I wanted to get some feedback on the overall feature first.

munificent:
I don't know if I'm sold on a user-visible option named `--undefok`. But I have fairly frequently wished that ArgParser was more flexible about unknown options. I never found the time to think it through, but I've wanted to add support for something like "dynamic" options.

We could maybe add something like:

```dart
class ArgParser {
  void handleUnknownOption(bool Function(String name, List<String> values) callback) { ... }
}
```

If you had that, then you could essentially implement `--undefok` in user code.

I'd rather something like that than having a specific option name baked into the class.

natebosch:
> I'd rather something like that than having a specific option name baked into the class.

That was my first thought too. In the end there are a few reasons I implemented it this way:
- We'd need other, potentially difficult, behavior differences to support this with a `handleUnknownOption` because we'd need to be sure that the callback for `--undefok` before the unknown options actually come in. I suppose that could be mitigated if we ensure that `handleUnkownOption` _does_ get called eagerly and the tools _don't_ manually add it as an option.
- We'll probably end up wanting to do this in all of our tools anyway and so we'd end up writing the same unkown option handler over and over.

We could minimize the boilerplate outside of this package I suppose...

```dart
extension Undefok on ArgParser {
  void useUndefokBehavior() {
    final allowed = <String>{};
    handleUnknownOption((name, values) {
      if (name == 'undefok') {
        allowed.addAll(values);
      } elseif (allowed.contains(name)) {
        // ignore
      } else {
        throw ArgParserException('Unknown option $name');
      }
    });
  }
}
```

But it still seems easier to define the behavior once and use it from all our tools.

munificent:
> * We'd need other, potentially difficult, behavior differences to support this with a `handleUnknownOption` because we'd need to be sure that the callback for `--undefok` before the unknown options actually come in. I suppose that could be mitigated if we ensure that `handleUnkownOption` _does_ get called eagerly and the tools _don't_ manually add it as an option.

Sorry, I didn't follow this. I think there's a word or two missing.

natebosch:
> Sorry, I didn't follow this. I think there's a word or two missing.

Yeah this was unclear - I was starting with the assumption that if we added an `handleUnknownOption` callback, it would work like the `callback` argument to `addFlag` and `addOption`. Specifically those callbacks _don't_ get called eagerly as arguments are parsed, and that is part of the reason  we can't implement `undefok` with the existing `callback` mechanism.

It's possible I'm overthinking this though - it should be feasible to do this entirely with post processing as long as we can suppress the exception that `args` wants to through, so I guess it really shouldn't matter if the callback is called eagerly.

mosuem:
Closing this, as it is stale and the files are out of date. Feel free to reopen in case you still want to land it!

PR #185, opened 2021-03-15
Split parse* methods to avoid duplicate work

natebosch:
Each of the methods `parseSoloOption`, `parseAbbreviation`, and
`parseLongOption` do some string manipulation and checking, before
potentially calling the same method on their parent parser. No state
changes during parsing part. Split out a 'handle' method in each case
that takes a partially parsed result as arguments, and avoid duplicating
that work at each step in the parent chain. There may be other work that
could be avoided but it would take deeper refactoring. The
`parseShortFlag` does not need to be split since there are no
intermediate results computed from the String.

Change all possible members of `Parser` to private.

### DIFF
--- a/lib/src/parser.dart
+++ b/lib/src/parser.dart
@@ -29,15 +29,14 @@ class Parser {
   final Queue<String> _args;
 
   /// The remaining non-option, non-command arguments.
-  final _rest = <String>[];
+  final List<String> _rest;
 
   /// The accumulated parsed options.
   final Map<String, dynamic> _results = <String, dynamic>{};
 
   Parser(this._commandName, this._grammar, this._args,
-      [this._parent, List<String>? rest]) {
-    if (rest != null) _rest.addAll(rest);
-  }
+      [this._parent, List<String>? rest])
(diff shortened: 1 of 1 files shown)


lib/src/parser.dart
```diff
@@ -16,57 +16,57 @@ import 'option.dart';
...
-  final Map<String, dynamic> results = <String, dynamic>{};
+  final Map<String, dynamic> _results = <String, dynamic>{};
 
-  Parser(this.commandName, this.grammar, this.args,
-      [this.parent, List<String>? rest]) {
-    if (rest != null) this.rest.addAll(rest);
+  Parser(this._commandName, this._grammar, this._args,
+      [this._parent, List<String>? rest]) {
+    if (rest != null) _rest.addAll(rest);
   }
```
munificent, line 40: While you're in here, maybe:

```dart
  final List<String> _rest;

  ...

  Parser(this._commandName, this._grammar, this._args,
      [this._parent, List<String>? rest]) : _rest = [...?rest];
```

munificent (approved):

Issue #194, opened 2021-05-04
Mandatory options aren't very useful


jamesderlin:
This was raised by https://stackoverflow.com/q/67384823/.

Currently the presence of a mandatory (https://github.com/dart-lang/args/pull/177) option will cause `ArgParser.parse` to throw an exception whenever the option is omitted.  I don't think that this is very useful in practice since it breaks `--help`.

https://github.com/dart-lang/args/issues/156 requested adding an equivalent of `required` options from Python's argparse module, but note that Python's argparse:
* Has a special `action='help'` mechanism for specifying that an option that should print help.
* Adds a special handler for `-h`/`--help` by default.

In contrast, `package:args`'s `ArgParser` currently is not opinionated about how help should be handled, and `-h`/`--help` is typically handled by an ordinary boolean flag.

PoloLacoste:
Hey !
I didn't think of that when I implemented it 😓.

I guess a simple message with all the missing required parameters and not an `Exception` should be better ?

PR #196, opened 2021-05-22
Fix bug with mandatory flag

PoloLacoste:
As mentioned in #194 there is a bug when you use the mandatory flag in an option with an help flag (to display the usage message). 

Example (from [here](https://stackoverflow.com/questions/67384823/how-to-use-mandatory-field-in-dart-args-package)):

```dart
final parser = ArgParser()
    ..addOption('name', abbr: 'n', help: 'Provide your name', mandatory: true)
    ..addFlag('help', abbr: 'h', help: 'Provide usage instruction', negatable: false);
final results = parser.parse(['--help']);
```
Result:
```
Unhandled exception:
FormatException: Option name is mandatory.
```
Instead we should see the usage message of the parser and then exit.

This pull request added a way to detect if there is a help flag and when used it will ignore the mandatory flags.
We can then use the parse results to check if an help flag has been parsed and display the help message.

```dart
if (results.wasParsed('help')){
  print(parser.usage);
  exit(0);
}
```

I think this is more a hotfix than a correct work around of the issue.

The first thing I tried to do was to print and then exit the `Parser`. But this is breaking quite a lot of tests...
Then I tried to remove the exit function and fix the problems I got with the `CommandRunner`.
I noticed that `CommandRunner` had a similar feature. When a help flag is detected it will display the usage message.
I was trying to do the same thing inside `ArgParser` but he is used inside `CommandRunner` and this is breaking some tests too.
To fix them I should know, inside the `ArgParser`, that we are used by a `CommandRunner` but I don't think this is a clean solution.

Maybe creating a `ProgramRunner` or `ArgsRunner` using the `ArgParser` to do like in the `CommandRunner` but with no commands and just simple args. 

If you have any advice on how to implement this kind of behavior, I would be pleased.

### DIFF
--- a/lib/src/arg_parser.dart
+++ b/lib/src/arg_parser.dart
@@ -290,10 +290,6 @@ class ArgParser {
     var allNames = [name, ...aliases];
 
     if (allNames.any((name) => findByNameOrAlias(name) != null)) {
-      // ignore duplicate help (needed to not break current lib or `test_core`)
-      if (name == 'help') {
-        return;
-      }
       throw ArgumentError('Duplicate option or alias "$name".');
     }
 
--- a/lib/src/arg_parser.dart
+++ b/lib/src/arg_parser.dart
@@ -119,9 +119,9 @@ class ArgParser {
   /// values from the [ArgResults].
   ///
   /// If [hide] is `true`, this option won't be included in [usage].
-  /// 
+  ///
   /// If [ignoreMandatory] is `true`, then passing any explicit value for this flag will cause
-  /// all other mandatory options to not be mandatory. 
+  /// all other mandatory options to not be mandatory.
   ///
   /// If [aliases] is provided, these are used as aliases for [name]. These
   /// aliases will not appear as keys in the [options] map.
--- a/lib/src/arg_parser.dart
+++ b/lib/src/arg_parser.dart
@@ -76,6 +76,7 @@ class ArgParser {
         _commands = commands,
         commands = UnmodifiableMapView(commands),
         allowTrailingOptions = allowTrailingOptions {
+          // register by default a help flag
           addFlag('help', abbr: 'h', negatable: false, help: 'Print this usage information.');
         }
 
@@ -293,7 +294,7 @@ class ArgParser {
 
     if (allNames.any((name) => findByNameOrAlias(name) != null)) {
 
-      // ignore duplicate help (needed to not break current lib, `test_core`)
+      // ignore duplicate help (needed to not break current lib or `test_core`)
       if(name == 'help') {
         return;
       }
--- a/lib/src/arg_parser.dart
+++ b/lib/src/arg_parser.dart
@@ -75,7 +75,9 @@ class ArgParser {
         options = UnmodifiableMapView(options),
         _commands = commands,
         commands = UnmodifiableMapView(commands),
-        allowTrailingOptions = allowTrailingOptions;
+        allowTrailingOptions = allowTrailingOptions {
+          addFlag('help', abbr: 'h', negatable: false, help: 'Print this usage information.');
+        }
 
   /// Defines a command.
   ///
@@ -288,7 +290,13 @@ class ArgParser {
       bool hide = false,
       List<String> aliases = const []}) {
     var allNames = [name, ...aliases];
+
     if (allNames.any((name) => findByNameOrAlias(name) != null)) {
+
+      // ignore duplicate help (needed to not break current lib, `test_core`)
+      if(name == 'help') {
+        return;
+      }
       throw ArgumentError('Duplicate option or alias "$name".');
     }
 
--- a/lib/src/arg_parser.dart
+++ b/lib/src/arg_parser.dart
@@ -134,6 +134,7 @@ class ArgParser {
       bool negatable = true,
       void Function(bool)? callback,
       bool hide = false,
+      bool ignoreMandatory = false,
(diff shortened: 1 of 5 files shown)


alextekartik:
What about other options that might not require a mandatory field - for example: `--version`

PoloLacoste:
I guess this is not the purpose of ArgParser to decide which option can bypass the mandatory check (but this is what I'm doing in this pull request)

I have two ideas two solve that:
- Adding a list of option names that will bypass the mandatory check.
- Adding a new parameter in an option, like `ignoreMandatory` to say that this option should bypass the mandatory check.

lib/src/arg_parser.dart
```diff
@@ -134,6 +134,7 @@ class ArgParser {
       bool negatable = true,
       void Function(bool)? callback,
       bool hide = false,
+      bool ignoreMandatory = false,
```
jakemac53, line 137: We need to document this on the method as well.

I would suggest some wording like:

```
If [ignoreMandatory] is `true`, then passing any explicit value for this flag will cause
all other mandatory options to not be mandatory. 
```

This brings up another question though, it is pretty weird to me that this takes effect whenever the flag is passed - regardless of the value passed for it.

In the case of `--help` and `--verbose`, we would want this to only happen if the value was `true` for instance.

lib/src/option.dart
```diff
@@ -20,13 +20,15 @@ Option newOption(
     bool? splitCommas,
     bool mandatory = false,
     bool hide = false,
+    bool ignoreMandatory = false,
```
jakemac53, line 23: I would honestly avoid even providing this option for options (only flags). I don't know of any existing use cases for options but I have a feeling any future use cases that might come up would only want to ignore the mandatory warnings if the option was provided a specific value, or something like that.

PoloLacoste: Actually a flag is an `Option`. This is the only structure I have to store any information.

PoloLacoste: Maybe creating a new structure like `Flag` would be a solution.

lib/src/arg_parser.dart
PoloLacoste: Thanks for the documentation, I had forgotten it.

I can add a new check to make sure, if there is a value passed to the flag, that it's `true`.

lib/src/option.dart
jakemac53: > Actually a flag is an `Option`. This is the only structure I have to store any information.

I understand that Flags are Options, but we don't have to expose the ability to pass this argument for options. You can only create flags/options through these methods, so we can avoid exposing anything for true options.

lib/src/arg_parser.dart
jakemac53: > I can add a new check to make sure, if there is a value passed to the flag, that it's `true`.

Ya, I do somewhat wonder if we should expose the ability to configure which value (true or false) should trigger it?

For instance we could change it to `bool? ignoreMandatoryIf`, and then you would either not set it at all, or pass an explicit boolean value which is the state you want to use to ignore mandatory options.

For instance now with the `true` check there is yet another bit of weirdness - what if you set `ignoreMandatory` on a flag whose default value is `true`? Now explicitly enabling it disables the checks but they aren't disabled by default, even though you haven't actually changed the values of any flags.

jakemac53: Another option could be to make `defaultsTo: true` and `ignoreMandatory: true` be incomatible

jakemac53:
cc @munificent @natebosch what are your thoughts on the api here?

lib/src/option.dart
PoloLacoste: Oh okay. Some maybe instead of the `newOption` function we could use a `newFlag` method.

jakemac53: Right, all we have to do is remove this argument from this function and then its only actually supported for flags, afaik at least :).

PoloLacoste: We can do the same thing for the option `negatable` right ?

lib/src/arg_parser.dart
PoloLacoste: > 
> 
> Another option could be to make `defaultsTo: true` and `ignoreMandatory: true` be incompatible

Yeah, this is definitively not two compatible options.

natebosch:
Another way we have handled this in the past is to handle `--help` without the arg parser.

```dart
void main(List<String> args) {
  if (args.contains('--help')) {
    // print usage, regardless of other args
  }
  // parse args normally
}
```

I think that gets tougher with commands. We could add a named argument to `parse` for this though?

```
void main(List<String> args) {
  final ignoreMandatory = !args.contains('--help');
  // initialize parser
  final results = parser.parse(args, ignoreMandatory: ignoreMandatory);
}
```

munificent:
This is starting to feel like the feature is adding complexity that requires more features to resolve, which adds complexity, which... etc.

How about we defer checking whether a mandatory option was passed until it is actually requested from the ArgResults object? This way, if the code never uses the option (because something like `--help` or `--version` was passed and the options are ignored), then no exception is thrown.

devoncarew:
@PoloLacoste - thanks for the PR here! I agree this part of `mandatory` options was a bit of a UX cliff - it stopped `-h` help working for commands that used `mandatory` (https://github.com/dart-lang/args/issues/245).

We landed a fix for this issue as part of #246 (based on the last recommendation in this PR).

PR #202, opened 2021-06-29
Add ability to group commands by category in usage

guidezpl:
To enable https://github.com/flutter/flutter/issues/83706

Formatting is loosely based on what `Brew` does, and open to suggestions.  Existing tests are unchanged.

Example (all displayed commands categorized):
Screenshot 2021-07-23 at 16 22 40

### DIFF
--- a/lib/command_runner.dart
+++ b/lib/command_runner.dart
@@ -436,6 +436,7 @@ String _getCommandUsage(Map<String, Command> commands,
   for (var category in categories) {
     if (category != '') {
       buffer.writeln();
+      buffer.writeln();
       buffer.write('$category');
     }
     for (var command in commandsByCategory[category]!) {
@@ -449,9 +450,6 @@ String _getCommandUsage(Map<String, Command> commands,
         buffer.write(' ' * columnStart);
         buffer.write(line);
       }
-    }
-    if (category != '' && category != categories.last) {
-      buffer.writeln();
     }
   }
 
--- a/lib/command_runner.dart
+++ b/lib/command_runner.dart
@@ -221,6 +221,12 @@ abstract class Command<T> {
   /// This defaults to the first line of [description].
   String get summary => description.split('\n').first;
 
+  /// The command's category.
+  ///
+  /// Displayed in [parent]'s [CommandRunner.usage]. Commands with categories
+  /// will be grouped together, and displayed after commands without a category.
+  String get category => '';
+
   /// A single-line template for how to invoke this command (e.g. `"pub get
   /// `package`"`).
   String get invocation {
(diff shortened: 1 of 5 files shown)


natebosch:
> We should at least validate that nobody is just implementing this class in the code we have access to (internal + dart owned packages + sdk?).

Yeah I think `CommandRunner.addCommand` already requires that you `extends Command` and not `implements Command`. Doubling down on that design flaw isn't a big problem IMO. It would be nice to do a redesign at some point, but I don't expect that will happen soon.

guidezpl:
Bump for review :)

lib/command_runner.dart
```diff
@@ -414,20 +420,43 @@ String _getCommandUsage(Map<String, Command> commands,
 
   // Show the commands alphabetically.
   names = names.toList()..sort();
+
+  // Group the commands by category.
+  var commandsByCategory = <String, List<Command>?>{};
```
jakemac53, line 425:
```suggestion
  var commandsByCategory = <String, List<Command>>{};
```

lib/command_runner.dart
```diff
@@ -414,20 +420,43 @@ String _getCommandUsage(Map<String, Command> commands,
...
+  for (var name in names) {
+    var category = commands[name]!.category;
+    if (commandsByCategory[category] == null) {
+      commandsByCategory[category] = [];
+    }
+    commandsByCategory[category]!.add(commands[name]!);
```
jakemac53, line 431:
```suggestion
    commandsByCategory.putIfAbsent(category, () => []).add(commands[name]!);
```

lib/command_runner.dart
```diff
@@ -414,20 +420,43 @@ String _getCommandUsage(Map<String, Command> commands,
...
+  }
+  // Sort categories alphabetically.
+  final categories = commandsByCategory.keys.toList();
+  categories.sort();
```
jakemac53, line 435:
```suggestion
  final categories = commandsByCategory.keys.toList()..sort();
```

lib/command_runner.dart
jakemac53: Nit: You could instead use a [SplayTreeMap](https://api.dart.dev/stable/2.13.4/dart-collection/SplayTreeMap-class.html) here and then you wouldn't have to sort the keys later on (it would iterate in sorted order by key).

test/command_runner_test.dart
```diff
@@ -50,6 +50,30 @@ Global options:
...
+A test command runner.
+
+Usage: test <command> [arguments]
+
+Global options:
+-h, --help    Print this usage information.
+
+Available commands:
+  foo   Set a value.
+Displayers
```
jakemac53, line 71: Is it intentional that there is no extra newline above this but there is an extra newline between the categories?

jakemac53 (commented):
We should add a few more tests as well, for instance:

- The case where all commands have categories
- Multiple commands in a single category

lib/command_runner.dart
guidezpl: Nice. Done

lib/command_runner.dart
guidezpl: Neat. Done

test/command_runner_test.dart
guidezpl: Yes, but it's a style choice I'll leave to you. Some other options:

Newlines 
```
Available commands:

Displayers
  baz   Display a value.

Printers
  bar   Print a value.
```

```
Available commands:
  foo   Set a value.

Displayers
  baz   Display a value.

Printers
  bar   Print a value.
```

Newlines + indent categories
```
Available commands:
  
  Displayers
  baz   Display a value.
  
  Printers
  bar   Print a value.
```

```
Available commands:
  foo   Set a value.
  
  Displayers
  baz   Display a value.
  
  Printers
  bar   Print a value.
```

No newlines 
```
Available commands:
Displayers
  baz   Display a value.
Printers
  bar   Print a value.
```

```
Available commands:
  foo   Set a value.
Displayers
  baz   Display a value.
Printers
  bar   Print a value.
```

jakemac53: I would personally do either no newlines or newlines always. In general I think the current style is pretty reasonable, but I do think a newline separating the categories from the uncategorized section would be good :).

I am not sure who all to seek opinion from here so ill just cc a few that I know 🤣  @natebosch @munificent @jonahwilliams

test/command_runner_test.dart
```diff
@@ -131,7 +179,9 @@ Run "test help <command>" for more information about a command.'''));
     });
 
     test("doesn't print hidden commands", () {
-      runner..addCommand(HiddenCommand())..addCommand(FooCommand());
+      runner
+        ..addCommand(HiddenCommand())
+        ..addCommand(FooCommand());
```
jakemac53, line 184: probably a good idea to add a hidden, categorized command in here as well?

Maybe also test that the category header is not displayed at all if all its commands are hidden?

test/command_runner_test.dart
jonahwilliams: I don't have any strong opinions here, though big +1 on this feature

test/command_runner_test.dart
guidezpl: Good idea, done.

test/command_runner_test.dart
guidezpl: Thanks! I went with always newlines

jakemac53 (approved):

jakemac53:
@googlebot I consent

jakemac53:
I will merge this into the Dart SDK and internally first, after which we can publish and roll into flutter

guidezpl:
Hey, wondering about the status for publishing & rolling

jakemac53:
Hi, apologies we just forgot to publish here, I can take a look today.

jakemac53:
Ok, this is now published. Thanks for the ping/reminder!

Flutter will pick this up the next time they update packages.

Issue #201, opened 2021-06-25
Not-found command should hint about similar ones


asashour:
As was mentioned in https://github.com/flutter/flutter/issues/74521

Shouldn't unknown commands handled like for e.g. `git`:

```
$ git alone
git: 'alone' is not a git command. See 'git --help'.

The most similar command is
        clone
```

jakemac53:
Ya, I have wanted this feature in args for a while 👍 . Some investigation is needed into which algorithm we should use, I know there are a few common ones for edit distance, but I don't have any domain expertise myself.

asashour:
I guess it doesn't really matter as a beginning, we can use [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) or [Dice coefficient](https://en.wikipedia.org/wiki/S%C3%B8rensen%E2%80%93Dice_coefficient), or even anything simpler.

jakemac53:
Ya its true that even a simple algorithm would likely help out a lot, supporting transposition intuitively seems like a nice improvement to me as well.

jakemac53:
We could even support transposition but only support suggestions with an edit distance of 1. That would be very efficient and easy to implement 🤷‍♂️ .

PR #206, opened 2021-07-08
show similar commands with an edit distance of 2 or less

jakemac53:
Fixes https://github.com/dart-lang/args/issues/201

Uses the [Optimal string alignment distance](https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance#Optimal_string_alignment_distance) algorithm to compute edit distance, which allows for additions, removals, substitutions, and swaps - all as 1 cost operations.

Suggests any commands within an edit distance of <=2, ~although we should possibly make this configurable~ which is now configurable.

We should also consider whether this should be opt in - it is likely to break existing integration tests in some cases. ~We could also consider adding in an opt out mechanism.~ you can now opt out by setting the distance to zero.

### DIFF
--- a/lib/command_runner.dart
+++ b/lib/command_runner.dart
@@ -473,8 +473,8 @@ String _getCommandUsage(Map<String, Command> commands,
 /// See https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance#Optimal_string_alignment_distance
 int _editDistance(String from, String to) {
   // Pad words with a space in front to mimic indexing by 1 instead of 0.
-  from = from.padLeft(1);
-  to = to.padLeft(1);
+  from = ' $from';
+  to = ' $to';
   var distances = [
     for (var i = 0; i < from.length; i++)
       [
--- a/lib/command_runner.dart
+++ b/lib/command_runner.dart
@@ -154,33 +154,6 @@ class CommandRunner<T> {
     Command? command;
     var commandString = executableName;
 
-    // Returns help text for commands similar to `name`, in sorted order.
-    String similarCommandsText(String name) {
-      if (suggestionDistanceLimit <= 0) return '';
-      var distances = <Command, int>{};
-      var candidates =
-          SplayTreeSet<Command<T>>((a, b) => distances[a]! - distances[b]!);
-      for (var command in commands.values) {
-        if (command.hidden) continue;
-        var distance = _editDistance(name, command.name);
-        if (distance <= suggestionDistanceLimit) {
-          distances[command] = distance;
-          candidates.add(command);
-        }
-      }
-      if (candidates.isEmpty) return '';
-      var similar = StringBuffer();
-      similar
-        ..writeln()
-        ..writeln()
-        ..writeln('Did you mean one of these?');
-      for (var command in candidates) {
-        similar.writeln('  ${command.name}');
-      }
-
-      return similar.toString();
-    }
-
     while (commands.isNotEmpty) {
       if (argResults.command == null) {
         if (argResults.rest.isEmpty) {
@@ -195,7 +168,8 @@ class CommandRunner<T> {
           var requested = argResults.rest[0];
 
           // Build up a help message containing similar commands, if found.
-          var similarCommands = similarCommandsText(requested);
+          var similarCommands =
+              _similarCommandsText(requested, commands.values);
 
           if (command == null) {
             usageException(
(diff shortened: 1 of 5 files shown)


jakemac53:
cc @jonahwilliams do you have opinions on this from a flutter tool perspective?

jakemac53:
cc @devoncarew @jwren for comment as well as this likely would be used for the `dart` cli (although it could be disabled)

jonahwilliams:
I love it

jakemac53:
One additional thing I realized here - this will currently suggest hidden commands. Should it do that?

jwren:
@bkonyi for `dartdev` fyi

LGTM.  Thanks for referencing the algorithm in the source code for future reference.

Nit: what about input where the user uses capitalization accidentally in the command name?

> hidden commands

Not sure.  If we do, we should write out that it is a hidden command: Did you mean `cmd_name (hidden)`?

jonahwilliams:
when commands are disabled in the flutter tool we normally don't even add them, so I don't think it will be an issue

bkonyi:
Awesome! I see no issues with this.

jakemac53:
> Nit: what about input where the user uses capitalization accidentally in the command name?

This should work fine for that - it wouldn't handle it specially so it would be an edit distance of 1 to correct.

lib/command_runner.dart
```diff
@@ -147,6 +154,21 @@ class CommandRunner<T> {
     Command? command;
     var commandString = executableName;
 
+    // Returns commands similar to `name`, in sorted order.
```
munificent, line 157: Does this need to be a local function? Can you make it a private method?

lib/command_runner.dart
```diff
@@ -147,6 +154,21 @@ class CommandRunner<T> {
...
+    Iterable<Command<T>> similarCommands(String name) {
+      final distances = Expando<int>();
```
munificent, line 159: Why is this an expando and not just:

```dart
var distances = <String, int>{};
```

?

(Also, we use `var` for all locals in args.)

lib/command_runner.dart
```diff
@@ -158,13 +180,32 @@ class CommandRunner<T> {
...
           command.usageException('Missing subcommand for "$commandString".');
         } else {
+          var requested = argResults.rest[0];
+
+          // Build up a help message containing similar commands, if found.
+          var similar = StringBuffer();
+          if (suggestedCommandsMaxEditDistance > 0) {
+            var possible = similarCommands(requested);
+            if (possible.isNotEmpty) {
+              similar.write(' The most similar command');
```
munificent, line 191: How about "Did you mean one of these?"

jakemac53:
I went ahead and changed this to omit hidden commands for now - we could always add a configuration option to enable it later

lib/command_runner.dart
```diff
@@ -77,7 +77,14 @@ class CommandRunner<T> {
   ArgParser get argParser => _argParser;
   final ArgParser _argParser;
 
-  CommandRunner(this.executableName, this.description, {int? usageLineLength})
+  /// The maximum edit distance allowed when suggesting possible intended
+  /// commands.
+  ///
+  /// Set to `0` in order to disable suggestions, defaults to `2`.
+  final int suggestedCommandsMaxEditDistance;
```
munificent, line 84: We're in `CommandRunner` so need to say "command" twice. How about just `suggestionDistanceLimit`?

lib/command_runner.dart
```diff
@@ -158,13 +180,32 @@ class CommandRunner<T> {
...
+              if (possible.length == 1) {
+                similar.writeln(' is:');
+              } else {
+                similar.writeln('s are:');
+              }
+              for (var command in possible) {
+                similar.writeln('  ${command.name}');
+              }
+            }
+          }
```
munificent, line 201: Also, I would move all this into the aforementioned method. Just have a single method that gives you the suggestion text and keep this one relatively simple.

lib/command_runner.dart
```diff
@@ -433,3 +474,41 @@ String _getCommandUsage(Map<String, Command> commands,
...
+  // Add a space in front to mimic indexing by 1 instead of 0.
+  from = ' $from';
+  to = ' $to';
+  var distances = [
+    for (var i = 0; i < from.length; i++)
+      [
+        for (var j = 0; j < to.length; j++)
+          if (i == 0) j else if (j == 0) i else 0,
+      ],
+  ];
```
munificent, line 494: Nit: Insert a blank line here and below the return statement.

lib/command_runner.dart
```diff
@@ -433,3 +474,41 @@ String _getCommandUsage(Map<String, Command> commands,
...
+  for (var i = 1; i < from.length; i++) {
+    for (var j = 1; j < to.length; j++) {
+      // Removals from `from`
```
munificent, line 497: "." at the end of all of these comments.

munificent (changes requested):
This is really cool. People are going to love it.

lib/command_runner.dart
jakemac53: This is the text that `git` uses, but I agree it feels a bit awkard. I will update it.

lib/command_runner.dart
jakemac53: It does read state from the surrounding scope. I could instead pass that as a parameter, but the local method makes it a bit cleaner and it isn't used outside this method?

lib/command_runner.dart
jakemac53: Updated, I also changed the formatting a bit, generally looks like this now:

```
Could not find a command named "$typo".

Did you mean one of these?
  long
```

jakemac53: cc @jonahwilliams @bkonyi how do you feel about the new text?

lib/command_runner.dart
munificent: If it were me, I would pass in the stuff as a parameter. This is quite long for a local function.

lib/command_runner.dart
```diff
@@ -492,24 +489,26 @@ int _editDistance(String from, String to) {
           if (i == 0) j else if (j == 0) i else 0,
       ],
   ];
+
   for (var i = 1; i < from.length; i++) {
     for (var j = 1; j < to.length; j++) {
-      // Removals from `from`
+      // Removals from `from`,
```
munificent, line 495: These should be ".", not ",". :)

https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

munificent (approved):

lib/command_runner.dart
jakemac53: lol whoops, I am typing with my left hand atm lol

munificent (approved):

Issue #203, opened 2021-06-30
Examples for CommandRunner aren't correct


mschwartz:
The examples show that you can use addFlags/addOption on the top level CommandRunner.  Yet there are no such methods on CommandRunner.

As well, I see nowhere in the code where global options are honored.

In the test suite, the tests do something like runner.argParser.addFlag(...), but again, those flags/options are not honored globally that I can see.

That is, a child of Command does not see these global flags/options in its argParser.options).

The workaround is to add your global options to each Command inheritor's constructor, which is a lot of duplicated effort and a bit of a maintenance headache - if you want to add a global option, you have to go edit all your *Command.dart files and add it to the constructors.

mschwartz:
I want to amend my issue with the following information.  There is a globalResults? in the Command class that does have the global flags values.  It's not well documented, IMO.

The examples are still incorrect :)

fengwensheng:
so now only can use like this below
`CommandRunner('CmdRunnerTester', 'test command runner')
    ..argParser.addFlag('verbose', abbr: 'v', help: 'show more detail...')
    ..addCommand(Cmd1())
    ..addCommand(Cmd2())
    ..run(args);`

bsutton:
Not quite certain if this related but I also can't find in the documentation a method to access the global args outside of a Command.

I'm wanting to parse the global arguments are a global level not within individual commands.
Is this a failing of the doco or is this missing functionality?

mschwartz:
Try looking at

command.globalResults

bsutton:
The problem is I'm trying parse the global commands at the global level and store the results in a factory that the commands other functions can then use.
So I want to parse the args and get the results before I run the commands.

I've worked around the problem by:
  var results = runner.argParser.parse(args);

but argParser is documented as a protected method.
I guess part of the issue is that I'm requiring that the args are parsed twice.

gresnick:
> so now only can use like this below `CommandRunner('CmdRunnerTester', 'test command runner') ..argParser.addFlag('verbose', abbr: 'v', help: 'show more detail...') ..addCommand(Cmd1()) ..addCommand(Cmd2()) ..run(args);`

This was helpful

```
import 'package:args/command_runner.dart';

CommandRunner cli =
      new CommandRunner(name, description)
        ..addCommand(new bootstrap.BootstrapCommand())
        ..addCommand(new diagnostics.DiagnosticsCommand())
        ..addCommand(new healthcheck.HealthCheckCommand())
[...]

PR #211, opened 2021-11-13
Update documentation

TheHackerCoding:
In order to add a global argument to `CommandRunner`, you need to add it through its `argParser` (reference to #203)


jakemac53 (approved):

jakemac53:
Thanks for the fix 🚀

PR #218, opened 2022-05-11
prep to publish 2.3.1

devoncarew:
prep to publish 2.3.1:
- rev tp 2.3.1
- update the changelog
- add markdown badges to the readme

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-## 2.3.1-dev
+## 2.3.1
+
+* Switch to using package:lints.
+* Address an issue with the readme API documentation (#211).
+* Populate the pubspec `repository` field.
 
 ## 2.3.0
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-## 2.3.1-dev
+## 2.3.1
+
+* Switch to using package:lints.
```
jakemac53, line 3: Fwiw we typically don't include this type of change in the changelog. It doesn't really hurt but it is probably just noise to consumers of the package.

jakemac53 (approved):

CHANGELOG.md
devoncarew: I'm mostly concerned with capturing the `repository` field change w/ a publish. It's a light patch release, so I included descriptions of the other changes in here as well.

jakemac53: It is an implementation detail of the package (does not affect consumers), and the changelog is generally for users. But, I don't feel strongly or anything about removing it, you can leave it in if you like.

Longer changelogs aren't better, if the info isn't useful :).

PR #220, opened 2022-08-04
Add a CommandRunner example

### LINKED ISSUE #219: Add a CommandRunner example
Since the `CommandRunner` API works by extending the `Command` class, it's not obvious from the API documentation how to use it (ie which methods are required/recommended to override). An example would be really helpful.

liamappelbe:
Uses `CommandRunner` to create a tool for drawing ascii art shapes.

Fixes #219


example/command_runner/draw.dart
```diff
@@ -0,0 +1,141 @@
...
+  @override
+  FutureOr<String>? run() {
+    final width = int.parse(argResults?['width'] ?? '50');
+    final height = int.parse(argResults?['height'] ?? '10');
+    final char = globalResults?['char']?[0] ?? '#';
+    return drawTriangle(width, height, char);
+  }
+}
+
+void main(List<String> args) async {
```
devoncarew, line 133: Perhaps move this method to the top of the file, so it's the first thing people see? You may want to sequence the commands next, and then the support methods (`draw`, `drawTriangle`, ...).

example/command_runner/draw.dart
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:math';
+import 'package:args/command_runner.dart';
```
devoncarew, line 7: nit: add a blank link between dart: and package: imports.

devoncarew (approved):
lgtm; two suggestions about ordering in the example and ws.

PR #235, opened 2023-02-06
Suggest command based on aliases

sigurdm:
Also introduce a concept of suggestionAliases (alternative names that cannot be used as a command name, but they will be used for suggesting the actual command).

The motivation comes from this suggestion: https://github.com/dart-lang/sdk/issues/50971#issuecomment-1412495274

That would allow us to give a good suggestion if someone writes `dart packages get` -> `Did you mean pub?`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.4.0
+## 2.4.0-dev
 
 * Command suggestions will now also suggest based on aliases of a command.
 * Introduce getter `Command.suggestionAliases` for names that cannot be used as
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.4.0
+
+* Command suggestions will now also suggest based on aliases of a command.
+* Introduce getter `Command.suggestionAliases` for names that cannot be used as
+  aliases, but will trigger suggestions.
+
 ## 2.3.2
 
 * Require Dart 2.18
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.4.0-dev
+## 2.4.0
 
 * Command suggestions will now also suggest based on aliases of a command.
 * Introduce getter `Command.suggestionAliases` for names that cannot be used as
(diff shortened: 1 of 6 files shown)


lib/command_runner.dart
```diff
@@ -412,6 +419,17 @@ abstract class Command<T> {
   /// This is intended to be overridden.
   List<String> get aliases => const [];
 
+  /// Alternate non-functional names for this command.
+  ///
+  /// These names won't be used in the documentation, and also they will work
```
jakemac53, line 424:
```suggestion
  /// These names won't be used in the documentation, and also they won't work
```

jakemac53 (commented):
This LGTM but we do need a pubspec/changelog update.

I also can't remember for certain whether it is supported to implement `Command` but I don't think so? It would be breaking if so, but I am fairly certain we always extend?

sigurdm:
>  It would be breaking if so, but I am fairly certain we always extend?

Good point about breakiness. I searched internally for usages, and the only place we `implement` this is in dart2js_info where it is used for a mixin. So in practice this seems quite non-breaking. However I don't think it is documented anywhere?

It seems we have precedence of adding properties to `Command` without a breaking release: https://github.com/dart-lang/args/pull/202.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.4.0-dev
```
jakemac53, line 1: Want to go right to a stable release? We don't have any other changes I know of coming.

jakemac53 (approved):

CHANGELOG.md
sigurdm: Ok with me.

jakemac53:
Filed https://github.com/dart-lang/ecosystem/issues/56 not sure what is up with that rpc error

devoncarew:
hmm, this:

```
RpcException: Forbidden
#0      Github.callRestApiPost.<anonymous closure> (package:firehose/src/github.dart:66:13)
<asynchronous suspension>
#1      Github.createComment (package:firehose/src/github.dart:104:18)
...
```

was - after a successful package validation - we tried to write a comment back to the PR saying something like `tag with xxx to invoke a publish`.

devoncarew:
Hmm, as a mea culpa, I published this without first making sure that it had rolled through the sdk and google3. I'll watch for any issues downstream.

Issue #95, opened 2018-02-24
ArgResults: Make easier for Dart2 users


matanlurey:
This is a runtime failure in Dart2:

```dart
void readListOfNames(ArgResults results) {
  // "List<dynamic> is not a List<String>"
  List<String> names = results['names'];
}
```

... but patterns like this are really common.

I'd like to see a series of helper methods to make it easier to get common types and collections out of ArgResults, even if all they do behind the scenes is a combination of `as` and collection casts. Ideas:

## Primitives

* [ ] `bool readBool(String name)`
* [ ] `double readDouble(String name)`
* [ ] `int readInt(String name)`
* [ ] `String readString(String name)`

## Collections

* [ ] `List<String> readStringList(String name)`

Thoughts?

/cc @nex3 @natebosch

nex3:
I think there's a general need for a good API to access heterogeneous untyped information. We have something like this for [JSON-RPC 2 parameters](https://www.dartdocs.org/documentation/json_rpc_2/2.0.7/json_rpc_2/Parameters-class.html), but it would be nice to generalize it enough that it could be used for argument values, JSON/YAML objects, and so on. As the pressures to avoid dynamic values get stronger, this will be more and more relevant.

srawlins:
The args package already has a delineation between bool "flags" and other-type "options." What about even just an `is` or `has` method for accessing bools?

```dart
var argParser = new ArgParser()
    ..addFlag('help');
var argResults = argParser.parse(args);
if (argResults['help']) { ... } // Error: Conditions must have a static type of 'bool'.
if (argResults['help'] == true) { ... } // What I have to do today. :(
if (argResults.has('help')) { ... } // Would be nice.
if (argResults.isNot('help')) { ... } // Lends itself nicely to negatable.
```

jonahwilliams:
I'm looking at disabling implicit downcasts in the flutter tool, and a large portion come from `ArgResults.[]` returning dynamic. I'm working around this with a typed wrapper which exposes `getOption`, `getFlag` and `getMultiOption` in https://github.com/flutter/flutter/pull/31679.

It would be great to have something similar in the args package.

lrhn:
You can do a lot with extensions these days.
```dart
extension ArgResultGetters on ArgResults {
   bool? optionalFlag(String name) => this[name] as bool?;
   bool flag(String name, {bool defaultValue: false}) => (this[name] as bool?) ?? defaultValue;
   bool has(String name) => this[name] != null;

[...]

PR #248, opened 2023-06-01
Introduce new typed `ArgResults` `flag(String)`, `option(String)`, and `multiOption(String)` methods

devoncarew:
- introduce new typed `ArgResults` `flag(String)`, `option(String)`, and `multiOption(String)` methods

This API lets people pull flags, options, and multi-options out of results objects without having to cast the values (e.g., `results['param1'] as bool`, `results['param2'] as String?`, ...). This is similar to the approach discussed in https://github.com/dart-lang/args/issues/95#issuecomment-487159630.

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>

### DIFF
--- a/lib/command_runner.dart
+++ b/lib/command_runner.dart
@@ -192,13 +192,13 @@ class CommandRunner<T> {
       commands = command._subcommands as Map<String, Command<T>>;
       commandString += ' ${argResults.name}';
 
-      if (argResults.options.contains('help') && argResults.getFlag('help')) {
+      if (argResults.options.contains('help') && argResults.flag('help')) {
         command.printUsage();
         return null;
       }
     }
 
-    if (topLevelResults.getFlag('help')) {
+    if (topLevelResults.flag('help')) {
       command!.printUsage();
       return null;
     }
--- a/lib/command_runner.dart
+++ b/lib/command_runner.dart
@@ -320,7 +320,7 @@ abstract class Command<T> {
   /// The parsed argument results for this command.
   ///
   /// This will be `null` until just before [Command.run] is called.
-  ArgResults get argResults => _argResults!;
+  ArgResults? get argResults => _argResults;
   ArgResults? _argResults;
 
   /// The argument parser for this command.
--- a/lib/command_runner.dart
+++ b/lib/command_runner.dart
@@ -192,13 +192,13 @@ class CommandRunner<T> {
       commands = command._subcommands as Map<String, Command<T>>;
       commandString += ' ${argResults.name}';
 
-      if (argResults.options.contains('help') && argResults['help']) {
+      if (argResults.options.contains('help') && argResults.getFlag('help')) {
         command.printUsage();
         return null;
       }
     }
 
-    if (topLevelResults['help']) {
+    if (topLevelResults.getFlag('help')) {
       command!.printUsage();
       return null;
     }
@@ -320,7 +320,7 @@ abstract class Command<T> {
   /// The parsed argument results for this command.
   ///
   /// This will be `null` until just before [Command.run] is called.
-  ArgResults? get argResults => _argResults;
+  ArgResults get argResults => _argResults!;
   ArgResults? _argResults;
 
   /// The argument parser for this command.
--- a/lib/src/arg_results.dart
+++ b/lib/src/arg_results.dart
@@ -122,7 +122,7 @@ class ArgResults {
   ///
   /// Includes the options whose values were parsed or that have defaults.
   /// Options that weren't present and have no default are omitted.
-  List<String> get options {
+  Iterable<String> get options {
     var result = _parsed.keys.toSet();
 
     // Include the options that have defaults.
@@ -130,7 +130,7 @@ class ArgResults {
       if (option.defaultsTo != null) result.add(name);
     });
 
-    return result.toList();
+    return result;
   }
 
   /// Returns `true` if the option with [name] was parsed from an actual
--- a/lib/src/arg_results.dart
+++ b/lib/src/arg_results.dart
@@ -62,7 +62,7 @@ class ArgResults {
   /// [name] must be a valid option name in the parser.
   ///
   /// > [!Note]
-  /// > Callers should prefer using the most strongly typed methods - [flag] for
+  /// > Callers should prefer using the more strongly typed methods - [flag] for
   /// > flags, [option] for options, and [multiOption] for multi-options.
   dynamic operator [](String name) {
     if (!_parser.options.containsKey(name)) {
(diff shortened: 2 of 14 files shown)


lib/command_runner.dart
```diff
@@ -320,7 +320,7 @@ abstract class Command<T> {
   /// The parsed argument results for this command.
   ///
   /// This will be `null` until just before [Command.run] is called.
```
devoncarew, line 322: Probably worth updating the docs here (if we keep this change). `'This will throw if called before...'`.

lib/command_runner.dart
```diff
@@ -320,7 +320,7 @@ abstract class Command<T> {
...
-  ArgResults? get argResults => _argResults;
+  ArgResults get argResults => _argResults!;
```
devoncarew, line 323: I suspect I should drop this change from the PR, and if we re-add it later, do something like - #224 - throw a StateError if this is null when called.

natebosch:
cc @jakemac53 @munificent

lib/src/arg_results.dart
```diff
@@ -73,6 +73,51 @@ class ArgResults {
...
 
+  /// Returns the parsed or default command-line flag named [name].
+  ///
+  /// [name] must be a valid flag name in the parser.
+  bool flag(String name) {
+    if (!_parser.options.containsKey(name)) {
+      throw ArgumentError('Could not find an option named "$name".');
+    }
+
+    var option = _parser.options[name]!;
```
jakemac53, line 84: I would assign this above, then check for null instead of the `containsKey` call. This will do one less map lookup and give you type promotion instead of having to rely on `!`.

lib/command_runner.dart
jakemac53: I would separate this out into a different PR at least - I don't have a strong opinion on the change though.

lib/src/arg_results.dart
```diff
@@ -57,7 +57,7 @@ class ArgResults {
       : rest = UnmodifiableListView(rest),
         arguments = UnmodifiableListView(arguments);
 
-  /// Returns the parsed ore default command-line option named [name].
+  /// Returns the parsed or default command-line option named [name].
   ///
```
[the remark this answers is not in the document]
munificent: I wouldn't mark it deprecated (or at least not yet), but I like the idea of having the doc comment here point to the new better methods.

test/parse_test.dart
```diff
@@ -83,6 +83,79 @@ void main() {
...
+
+      test('throws if requested as a multi-option', () {
+        var parser = ArgParser();
+        parser.addFlag('a', defaultsTo: true);
+        var results = parser.parse(['--a']);
+        throwsIllegalArg(() => results.multiOption('a'));
+      });
+    });
+
+    group('flag', () {
```
munificent, line 95: Nit: I would do `group('flag()', () { ... })` to match other groups and make it clear this is about that method.

test/parse_test.dart
```diff
@@ -83,6 +83,79 @@ void main() {
...
+      test('are true if present', () {
```
munificent, line 96: Nit: I would work this so that it reads like a phrase when prepended with the surrounding `group()` description, so "returns true if present". That way if the test fails, you'll see a failure on "flag() returns true if present". Likewise below.

munificent (approved):
This looks great!

lib/src/arg_results.dart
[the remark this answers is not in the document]
devoncarew: Yeah, I was a little hesitant about deprecating it given that essentially every user of this package will get deprecation messages.

Updating the docs here make sense.

test/parse_test.dart
devoncarew: sgtm - done

lib/src/arg_results.dart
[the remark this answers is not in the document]
devoncarew:
```
  /// > [!Note]
  /// > Callers should prefer using the most strongly typed methods - [flag] for
  /// > flags, [option] for options, and [multiOption] for multi-options.
```

devoncarew:
Hmm, this PR has introduced a bit of asymmetry in the API. With the new `bool flag(String)` ArgResults method, you can only retrieve a non-null bool from the parsed results.

However, with the existing `ArgParser.addFlag()` method, you can set a null default value for the flag (`bool? defaultsTo = false`):

https://github.com/dart-lang/args/blob/master/lib/src/arg_parser.dart#L129C3-L136C42

So an ArgParser can be created that will always throw when you try and retrieve the flag value from the ArgResults.

I could see a few resolutions here:

1) update the `ArgParser.addFlag` method to not allow a null value for `defaultsTo`; rev the major version of package:args to capture the API change.
2) ignore this issue for now; for places where people do want to pull out a null value, let them rely on the existing `argResults[flagName]` mechanism. Create an issue to remove the nullability with the next major package publish.

I'd lean towards option 2 above.

natebosch:
I suppose folks who are broken by the change can move to checking `wasParsed` to handle whatever behavior they intended from the `null` default? Option 2 SGTM.

devoncarew:
Yeah, I ran across this when updating the dart cli to use these new methods. `dart compile wasm` uses a strange tri-state for a `--minify` flag - `null`, `true`, or `false`. I updated it to use wasParsed (`args.wasParsed('minify') ? null : args.flag('minify')`), but really don't think the pattern's that great - supporting a null default value.

munificent:
You could add something like `bool? ArgResults.optionalFlag()` to check the tri-state value of a nullable flag.

natebosch:
> You could add something like `bool? ArgResults.optionalFlag()` to check the tri-state value of a nullable flag.

I think I'm more inclined not add this in favor of keeping the undesirable pattern localized to where it is used. If code is doing something usual like a tri-state flag, it's OK for the code to also look unusual and need an extra `wasParsed` check.

If some project has strong opinions about the API they can hide that pattern behind an `optionalFlag` extension method.

Issue #240, opened 2023-04-07
[Feature request] Show command alias in usage


CaiJingLong:
`pnpm` will show the alias of the command in usage, can we add it?

<details><summary>pnpm</summary>
<p>

```shell
Version 8.0.0
Usage: pnpm [command] [flags]
       pnpm [ -h | --help | -v | --version ]

Manage your dependencies:
      add                  Installs a package and any packages that it depends on. By default, any new package is installed as a prod dependency
      import               Generates a pnpm-lock.yaml from an npm package-lock.json (or npm-shrinkwrap.json) file
   i, install              Install all dependencies for a project
  it, install-test         Runs a pnpm install followed immediately by a pnpm test
  ln, link                 Connect the local project to another one
      prune                Removes extraneous packages
  rb, rebuild              Rebuild a package
  rm, remove               Removes packages from node_modules and from the project's package.json
      unlink               Unlinks a package. Like yarn unlink but pnpm re-installs the dependency after removing the external link
  up, update               Updates packages to their latest version based on the specified range

Review your dependencies:
      audit                Checks for known security issues with the installed packages
      licenses             Check licenses in consumed packages
  ls, list                 Print all the versions of packages that are installed, as well as their dependencies, in a tree-structure
      outdated             Check for outdated packages

Run your scripts:
      exec                 Executes a shell command in scope of a project
      run                  Runs a defined package script
      start                Runs an arbitrary command specified in the package's "start" property of its "scripts" object
   t, test                 Runs a package's "test" script, if one was provided

Other:
      pack
      publish              Publishes a package to the registry
      root

Manage your store:
      store add            Adds new packages to the pnpm store directly. Does not modify any projects or files outside the store
      store path           Prints the path to the active store directory
      store prune          Removes unreferenced (extraneous, orphan) packages from the store
      store status         Checks for modified packages in the store

Options:
  -r, --recursive          Run the command for each project in the workspace.

``` 

</p>
</details>

CaiJingLong:
Similar, but this is an alias for a command, not an option.
For example, a typical pub has a publish command whose alias is lish
But I can't see it from anywhere with -h.
I think -h as a usage should not hide this shorthand.

<img width="562" alt="image" src="https://user-images.githubusercontent.com/14145407/231406063-560a3971-aa7d-477a-b8ee-be56883f01e0.png">

gkc:
I had a need to show option aliases in the usage output so I've made a draft pull request from my fork which
[...]

PR #262, opened 2024-01-24
feat: show aliases in usage

gkc:
- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

**What I did**
- Add a flag to allow option alias(es) to be output by ArgParser's usage
- The default behaviour remains the same - i.e. aliases are not printed in usage
- Kind of like what is requested in https://github.com/dart-lang/args/issues/240 but for ArgParsers' Options, not CommandRunners' Commands

**How I did it**
- usage_test.dart:
  - Add a test to verify that aliases are shown for options if requested
  - Add a test to verify that aliases are not shown for options by default
- usage.dart: make _longOption also show aliases, if any, if showAliases is true
- added showAliases parameters in various places so that we can preserve current behaviour as default
- updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.4.3
+
+* Show alias(es) in usage
+
 ## 2.4.2
 
 * Change the validation of `mandatory` options; they now perform validation when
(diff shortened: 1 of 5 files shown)


mosuem:
Closing as the [dart-lang/args](https://github.com/dart-lang/args) repository is merged into the [dart-lang/core](https://github.com/dart-lang/core) monorepo. Please re-open this PR there!

PR #270, opened 2024-04-17
Update to latest lints, require Dart 3.3

kevmoo:


.github/workflows/publish.yaml
```diff
@@ -4,7 +4,7 @@ name: Publish
 
 on:
   pull_request:
-    branches: [ master ]
+    branches: [ main ]
```
devoncarew, line 7: 👍

pubspec.yaml
```diff
@@ -9,8 +9,8 @@ topics:
   - cli
 
 environment:
-  sdk: ^3.0.0
+  sdk: ^3.3.0
```
devoncarew, line 12: Do we really need as late as 3.3?

devoncarew (approved):

pubspec.yaml
kevmoo: In my mind: it hurts NOTHING. Makes sure we're getting access to latest language features. Given how quickly our ecosystem moves forward, etc.

